PR #12, opened 2020-12-09
Thibault/category

thibmonsel:


src/Category.cpp
```diff
@@ -0,0 +1,60 @@
...
+}
+
+std::ostream & operator <<( std::ostream & out, const Category & cat ){
+    /**
+    * Overloading << operator for Category class
+    *
+    * @param set_rank 
+    * 
+    */
+    out << "Category(" << cat.category_id_ << ":" << cat.rank_ << ")";
```
Mostah: Should we add a space between ':' and cat.rank? Might make it a bit more readable

test/TestCategory.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+  Category catCopied = Category(category1);
+
+  std::ostringstream os;
+  os << catCopied;
+  EXPECT_EQ(os.str(), "Category(cat1:13)");
+}
+
+TEST(TestCategory, TestAllInstancesDestroyed) {
+  EXPECT_EQ(AtomicMCDAObject::get_nb_instances(), 0);
```
Mostah, line 33: Not 100% sure how it work but I added `int AtomicMCDAObject::nb_instances_{0};` at the begining of my file, I think we need that because if there are other destructors failing (and thus nb_instances not at 0) it will also fail on that test and make things a bit confusing

Mostah (approved):
After fixing the issue with the tests, it should be good to merge 👍

src/Category.cpp
thibmonsel: Make that change

test/TestCategory.cpp
thibmonsel: added AtomicMCDAObject nb instance in the TestCategory

PR #14, opened 2020-12-10
Thomas/criteria

Mostah:


thibmonsel (approved):

thibmonsel (approved):

thibmonsel (approved):

src/Criteria.cpp
```diff
@@ -0,0 +1,81 @@
+#include "Criteria.h"
+#include "Criterion.h"
+#include <iostream>
+#include <string>
+#include <vector>
+
+Criteria::Criteria(std::vector<Criterion> &criterion_vect) {
+  // deep copy
+  for (int i = 0; i < criterion_vect.size(); i++) {
+    criterion_vect_.push_back(Criterion(criterion_vect[i]));
```
edix9, line 10: why do we need to convert to Criterion criterion_vect[i] here?

test/TestCriteria.cpp
```diff
@@ -0,0 +1,124 @@
+#include "../src/Criteria.h"
+#include "../src/Criterion.h"
+#include "gtest/gtest.h"
+#include <sstream>
+#include <utility>
+
+TEST(TestCriteria, TestBaseConstructorWithNbAndId) {
+  std::string id = "test";
+  Criteria crits = Criteria(2, id);
```
edix9, line 9: we should be careful here with variables name ("id" here is used for another purpose in the next lines)

edix9 (approved):
looks good

test/TestCriteria.cpp
thibmonsel: what do you mean ? 
id is just for the testing.

PR #16, opened 2020-12-10
Thibault/category

thibmonsel:


src/Categories.h
```diff
@@ -0,0 +1,113 @@
...
+   */
+  Categories(int number_of_categories, std::string prefix);
+
+  /**
+   * Categories standard constructor 2
+   *
+   * @param number_of_categories number of categories wanted
+   */
+  Categories(int number_of_categories);
+
```
Mostah, line 32: Would it be usefull as well to add a constructor where we create a Categories object from a vect of category ?

src/Categories.h
```diff
@@ -0,0 +1,113 @@
...
+                                  const Categories &categories);
+
+  /**
+   * Overloading [] operator for Categories
+   *
+   * @param index index of the object we want
+   *
+   *@return Category object at index position of Categories object
+   */
+  Category operator[](int index);
```
Mostah, line 104: It's a nice feature 👍  would be use to add it to the other types as well!

src/Categories.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+Category Categories::operator[](int index) { return categories_vector_[index]; }
+
+Category Categories::operator[](int index) const {
+  return categories_vector_[index];
+}
+
+Categories::~Categories() {
+  // https://stackoverflow.com/questions/993590/should-i-delete-vectorstring
+  std::vector<Category>().swap(categories_vector_);
```
Mostah, line 46: If that's absolutely necessary, we need to add it to the other classes as well !

Mostah (approved):

src/Categories.h
thibmonsel: Added a new constructor

PR #18, opened 2020-12-14
Thomas/base perf classes

Mostah:


src/Perf.h
```diff
@@ -0,0 +1,78 @@
...
+  /**
+   * Perf standard constructor
+   *
+   * @param name name of the performance (ex: name of profile or alternative)
+   * @param criterion criterion id on which the performance in based on
+   * @param value value of the performance
+   */
+  Perf(std::string name, std::string criterion, float value);
+
+  /**
```
thibmonsel, line 19: Maybe we could add  new constructors for example : 
- build a Perf object with only name
- build a Perf object with only criterion
- build a Perf object with name and criterion
- build a Perf with only its value 
I think the third constructor might come in handy, the others they might but not sure ....

src/Perf.h
```diff
@@ -0,0 +1,78 @@
...
+class Perf : public AtomicMCDAObject {
+public:
+  /**
+   * Perf standard constructor
+   *
+   * @param name name of the performance (ex: name of profile or alternative)
+   * @param criterion criterion id on which the performance in based on
+   * @param value value of the performance
+   */
+  Perf(std::string name, std::string criterion, float value);
```
thibmonsel, line 17: What is better float or doubles for precision and memory allocation ?

src/Performance.h
```diff
@@ -0,0 +1,60 @@
...
+#include <vector>
+
+class Performance {
+public:
+  /**
+   * Performance standard constructor
+   *
+   * @param id Id name of the performance (ex: name of alternative or profile)
+   * @param criteria Array of criterion to compute performance over
+   */
```
thibmonsel: Is it a std::array or std::vector ?

src/Performance.h
```diff
@@ -0,0 +1,60 @@
...
+  Performance(std::string id, Criteria &criteria);
+
+  /**
+   * Performance standard constructor
+   *
+   * @param id Id name of the performance (ex: name of alternative or profile)
+   * @param criteria Array of criterion to compute performance over
+   * @param performance Array of performance (float) to set
+   */
+  Performance(std::string id, Criteria &criteria, std::vector<float> &p);
```
thibmonsel: Is it a std::array or std::vector ? 
The arguments suggests vector and not arrays

src/Performance.h
```diff
@@ -0,0 +1,60 @@
...
+
+  friend std::ostream &operator<<(std::ostream &out, const Performance &p);
+
+  /**
+   * getPerf getter of weight parameter
+   *
+   * @return perf
+   */
+  std::vector<Perf> getPerf() const;
+
```
thibmonsel: Not sure why we are talking about the weight parameter here.
If I'm not mistaken, the weight parameter only plays in for the Criteria object and not the performances of alternatives or profiles. What we would be taking about here, would be to get the evaluation of the performance on a given criteria.
ie if performance = [(a1,c1,0.1), (a1,c2,0.43), (a1,c3, 0.54)]
the method getPerf() would probably need an integer argument or criterion argument to indicate which value of Perf object you want.

src/PerformanceTable.h
```diff
@@ -0,0 +1,60 @@
...
+
+class PerformanceTable {
+public:
+  /**
+   * Performances constructor with defined vector of performance
+   *
+   * @param perf_vect Vector of performance
+   */
+  PerformanceTable(std::vector<Performance> &perf_vect);
+
```
thibmonsel, line 17: Will we get a PT that will have a sort of matrix evaluation ? 
ie could we access pt[i][j] which is the performance of the j_th criteria of the i_th alternative ?
I don't if we can do that easily with this structure or we might need to create an [][] operator overload ?

src/Perf.h
LaurentCabaret: Precision --> double
memory & vectorisation --> float
I asked Vincent  in June : float are more than enough for the precision

src/Perf.h
Mostah: I will add the third one with default value of 0 but the others are not likely to be useful in our case: we need to know which criterion is used and from where is the data from (name parameter)

src/Performance.h
Mostah: It's a std::vector, I'll update the doc

src/Performance.h
Mostah: It's a std::vector, I'll update the doc

src/Performance.h
Mostah: That's a copy paste mistake, we are talking about a std::vector<Perf> here, not a weight parameter. I'll add a [] operator (or similar) to get data from a specific criteria

src/PerformanceTable.h
Mostah: Yes, I'll add a [] operator to the PT to get a direct acccess to a Performance vector, thus the [i][j] operator will access a performance of the j_th criteria of the i_th alternative, at a cost of O(n^2) given the implementation. However this huge access cost is not likely to be an issue in our case as it will not be a common case to access directly a Perf object from the PT.

thibmonsel: To implement the get_best, get_worst, get_middle, etc ... we will need to get the each column of the PT so we will need to access those elements so we will have a cost of O(n^2) ? I that's the case ( I think so), might need to think about other data types ?

Mostah: The datastructure was designed to be optimized to be itterated over all its content (for sorting algorithms for example), as it will not be a common case to access a single Perf directly. get_above, get_best and get_middle are particular functions that will be implemented in O(1) as we already know where the elements will be (first position, last position, middle position). If at some point accessing the value of a specific criteria of a specific alternative (pt[i][j]) becomes a common operation we will need to modify the datastructure to get an access time in O(1). However I haven't thought of any ordered datastructure with O(1) access time in C++, feel free if you have suggestions

thibmonsel (approved):

PR #19, opened 2020-12-16
Thomas/sorted pt

Mostah:


src/Performance.h
```diff
@@ -65,6 +65,14 @@ class Performance {
...
 
+  /**
+   * getCriteria get of the criteria name used in the performance
+   *
+   * @return crit_vect string vector containing the criterion names used in this
+   * performance.
+   */
+  std::vector<std::string> getCriteria();
+
   /**
```
thibmonsel: change the method function to get CriteriaId() to remove ambiguity

src/PerformanceTable.cpp
```diff
@@ -1,49 +1,230 @@
...
+    std::sort(pv.begin(), pv.end(), [](const Perf &a, const Perf &b) {
+      return a.getValue() < b.getValue();
+    });
+  }
+}
+
+std::vector<Perf> PerformanceTable::getAltBetween(std::string crit, float inf,
+                                                  float sup) {
+  if (mode_ != "crit") {
+    throw std::domain_error("Mode must be crit but is currently set to " +
```
thibmonsel, line 212: Lets call the argument std::string with a more meaningful name ie are we talking about criterion, criteria and what ? (ie criterion id, criterion name etc ...)

thibmonsel (approved):

src/PerformanceTable.cpp
```diff
@@ -1,49 +1,306 @@
...
+void PerformanceTable::generateRandomPerfValues(unsigned long int seed) {
+  srand(seed);
+  for (std::vector<Perf> &pv : pt_) {
+    for (Perf &p : pv) {
+      p.setValue((float)rand() / RAND_MAX);
+    }
+  }
+  sorted_ = false;
+}
+
```
thibmonsel: should we use a have a lower and upper bound limit for the values specified ? 
ie by default its between 0 and 1 but you can give btwn 1 and 4 for example ?

thibmonsel (approved):

src/PerformanceTable.cpp
Mostah: Yeah we could add this, but aren't the values supposed to be in [0,1] ?

thibmonsel: I think they can be as big as one want but for optimization purpose the data probably needs to be normalised

Mostah: I see, we will need to add these normalization functions! I will add some bound limit args here to be more flexible

src/PerformanceTable.h
```diff
@@ -19,10 +21,11 @@ class PerformanceTable {
    * Criteria constructor without perf values but set of criteria to evaluate
```
edix9: Performances instead of Criteria?

PR #22, opened 2020-12-17
Emma/generate methods

edix9:


src/Criterion.cpp
```diff
@@ -24,6 +24,19 @@ Criterion::Criterion(const Criterion &crit){
...
+    else { direction_ = 1; }
+}
+
+void Criterion::generateWeight(unsigned long int seed){
+    srand(seed);
+    weight_ = (float) rand() / RAND_MAX ;
+}
+
+
+
```
thibmonsel, line 39: Should we have one extra function that combines both in order to generate a random Criterion with one method function call ?

Mostah: We could implement such functions indeed, maybe let's wait to analyze the different algorithms to check which initializations are we going to use?

src/Criterion.h
```diff
@@ -37,6 +37,21 @@ class Criterion : public AtomicMCDAObject  {
     ~Criterion();
     friend std::ostream & operator <<( std::ostream & out, const Criterion & crit );
 
+    /**
+     * generateDirection generator of random direction
+     * can also be used as reinitialisation for direction
+     *
```
Mostah: nit: for the doc I think it's better if we add the params in its description (with @param)

src/Criterion.h
```diff
@@ -37,6 +37,21 @@ class Criterion : public AtomicMCDAObject  {
...
+    /**
+     * generateDirection generator of random direction
+     * can also be used as reinitialisation for direction
+     * 
+     */
+    void generateDirection(unsigned long int seed = time(NULL));
+
+    /**
+     * generateWeight generator of random weight
+     * can also be used as reinitialisation for weight
```
Mostah: nit: same here

src/Criterion.cpp
thibmonsel: That function would be useful for the dataGenerator im working on for example 👍

thibmonsel: and where we could control the randomness with a seed in order to make test work

test/TestCriterion.cpp
```diff
@@ -36,6 +36,78 @@ TEST(TestCriterion, TestConstructorByCopy) {
...
+  Criterion crit2 = Criterion(id2);
+  unsigned long int seed2 = 40;
+  // generating a direction to compare to the one generated in generateDirection
+  // should be the same since we are using the same seed 
+  srand(seed);
+  int direction = 1;
+  if ( ((float) rand() / RAND_MAX) < 0.5){ direction = -1; }
+  crit.generateDirection(seed);
+
+  srand(seed2);
```
Mostah: It looks like you are testing twice the same feature here

test/TestCriterion.cpp
```diff
@@ -36,6 +36,78 @@ TEST(TestCriterion, TestConstructorByCopy) {
...
+  // should be the same since we are using the same seed 
+  srand(seed); 
+  float weight = ((float) rand() / RAND_MAX);
+  crit.generateWeight(seed);
+
+  srand(seed2); 
+  float weight2 = ((float) rand() / RAND_MAX);
+  crit2.generateWeight(seed2);
+
+  float weight3 = ((float) rand() / RAND_MAX);
```
Mostah: Looks like you could remove `weight3` as you are not using it

test/TestCriterion.cpp
```diff
@@ -36,6 +36,78 @@ TEST(TestCriterion, TestConstructorByCopy) {
...
+  
+  EXPECT_FLOAT_EQ(weight, crit.getWeight());
+  EXPECT_FLOAT_EQ(weight2, crit2.getWeight());
```
Mostah: same comment as for the test above, looks like you are testing twice the same feature

Mostah (approved):
A few comments to clean up a bit the code, otherwise it looks good to me 👍

src/Criterion.cpp
edix9: will be solved in Thibault's generator

src/Criterion.h
edix9: corrected

src/Criterion.h
edix9: corrected

test/TestCriterion.cpp
edix9: removed

test/TestCriterion.cpp
edix9: Right, I just wanted to test with different seeds but not really useful. Will remove

PR #23, opened 2020-12-17
Thibault/dataset generator

thibmonsel:


src/Criterion.cpp
```diff
@@ -2,36 +2,34 @@
...
+  direction_ = 1;
+  weight_ = 0.;
+}
 
-Criterion::Criterion(std::string id){
-    id_ = id;
-    name_ = "";
-    direction_ = 1;
-    weight_ = 0.;
+Criterion::Criterion(std::string id, int direction, float weight) {
```
Mostah, line 6: I think you could reduce this constructor and the previous one into just one with `direction` and `weight` as optional

src/Criterion.cpp
```diff
@@ -2,36 +2,34 @@
...
+  direction_ = crit.getDirection();
+  weight_ = crit.getWeight();
 }
 
-Criterion::Criterion(const Criterion &crit){ 
-    id_ = crit.getId();
-    name_ = crit.getName();
-    direction_ = crit.getDirection();
-    weight_ = crit.getWeight();
+Criterion::Criterion(std::string id, float weight) {
```
Mostah: Same comment as above, we could get rid of this one too

src/Criterion.cpp
```diff
@@ -40,19 +38,16 @@ float Criterion::getWeight() const { return weight_; }
...
-    return out;
+std::ostream &operator<<(std::ostream &out, const Criterion &crit) {
+  std::string dir = "";
+  if (crit.direction_ == 1) {
+    dir = "+";
+  } else {
+    dir = "-";
+  }
+  out << "Criterion(id : " << crit.id_ << ", direction : " << dir
+      << ", weight : " << crit.weight_ << ")";
```
Mostah, line 59: By changing the `<<` operator, it's probably going to fail some tests in the TestCriterion file, you might need to add the changes as well in this PR for the pipelines to pass

src/DataGenerator.cpp
```diff
@@ -0,0 +1,499 @@
...
+#include <fstream>
+#include <iostream>
+#include <sstream>
+#include <stdlib.h>
+#include <string>
+
+/// room for improvement can i improve the xml type fucntion by returning doc
+/// type
+
+void DataGenerator::DatasetGenerator(int nb_criteria, int nb_alternative,
```
Mostah: I think it's as better practice to put the default value of a param in the header file

src/DataGenerator.cpp
```diff
@@ -0,0 +1,499 @@
...
+  // Creating root
+  auto root = doc.append_child("Root");
+  pugi::xml_node dataset_node = doc.append_child("dataset");
+
+  // Giving dataset name
+  pugi::xml_node model_name = dataset_node.append_child("datasetName");
+  std::string name = "dataset_alt" + std::to_string(nb_alternative) + "_crit" +
+                     std::to_string(nb_criteria) + "_cat" +
+                     std::to_string(nb_categories);
+
```
Mostah, line 32: Should we handle the case where an already existant dataset exists with the given name (or the generated one)?

src/DataGenerator.cpp
```diff
@@ -0,0 +1,499 @@
...
+  for (int i = 0; i < nb_alternative; i++) {
+    pugi::xml_node alternative_node = dataset_node.append_child("alternative");
+
+    Performance alternative = Performance("alt" + std::to_string(i), criterias);
+
+    // getting alternative id
+    alternative_node.append_child(pugi::node_pcdata)
+        .set_value(alternative.getId().c_str());
+
+    for (int i = 0; i < alternative.getPerf().size(); i++) {
```
Mostah: to simplify the for loop here, instead of iterating with `i` you can also use this synthax:
```
for( Perf perf : alternative.getPerf()) {
     perf.getCrit()
     ...
```

I find it a bit easier to read :)

But you don't have to make the changes here, just in case you didn't know it!

src/DataGenerator.cpp
```diff
@@ -0,0 +1,499 @@
...
+    // Give the category assignemnt to the alternative
+    pugi::xml_node alternative_assignment =
+        alternative_node.append_child("assignment");
+
+    alternative_assignment.append_child(pugi::node_pcdata)
+        .set_value(
+            "need to get category for it modify performance or other types");
+  }
+
+  if (datasetName == "") {
```
Mostah, line 86: About my comment above L33, you could do the handling for the name here as well by making sure we catch the exception

src/DataGenerator.cpp
```diff
@@ -0,0 +1,499 @@
...
+    std::string modelpath = "../data/" + name + ".xml";
+  }
+  std::string modelpath = "../data/" + datasetName;
+
+  bool saveSucceeded = doc.save_file(modelpath.c_str(), PUGIXML_TEXT("  "));
+  assert(saveSucceeded);
```
Mostah: It might be better to raise an exception in case of a failure here instead of an assert, the more verbose, the better!

src/DataGenerator.cpp
```diff
@@ -0,0 +1,499 @@
...
+      .set_value(std::to_string(nb_criteria).c_str());
+
+  // Giving number of categories in total
+  pugi::xml_node categories_node = dataset_node.append_child("categories");
+  categories_node.append_child(pugi::node_pcdata)
+      .set_value(std::to_string(nb_categories).c_str());
+
+  // Giving lambda
+  srand((unsigned)time(NULL));
+  float lambda = (float)(rand() % 1000) / 2000 + 0.5;
```
Mostah: Why did you use this expression for lambda? Can you add some comments on that to be sure we all understand that value?

src/DataGenerator.cpp
```diff
@@ -0,0 +1,499 @@
...
+        .set_value(
+            std::to_string(criteria.getCriterionVect()[i].getWeight()).c_str());
+
+    if (modelName == "") {
+      std::string modelpath = "../data/" + name + ".xml";
+    }
+    std::string modelpath = "../data/" + modelName;
+
+    bool saveSucceeded = doc.save_file(modelpath.c_str(), PUGIXML_TEXT("  "));
+    assert(saveSucceeded);
```
Mostah: Same comment as for the DatasetGenerator about the assert and the handle of the fileName

src/DataGenerator.cpp
```diff
@@ -0,0 +1,499 @@
...
+  pugi::xml_document doc;
+  std::string path = "../data/" + fileName;
+  if (!doc.load_file(path.c_str()))
+    throw std::invalid_argument("Cannot open xml file, please check path");
+
+  // if we have a model xml
+  pugi::xml_node node = doc.child("model").child("modelName");
+
+  if (*node.child_value()) {
+    // if we have a dataset xml
```
Mostah, line 341: There's something misleading here between the comment and the return

src/DataGenerator.cpp
```diff
@@ -0,0 +1,499 @@
...
+    throw std::invalid_argument("Cannot open xml file, please check path");
+
+  if (DataGenerator::getXmlFileType(fileName) == "model") {
+    // if we have a model xml
+    pugi::xml_node lambda_node = doc.child("model").child("lambda");
+    float lambda = atof(lambda_node.child_value());
+    return lambda;
+  } else {
+    // if we have a dataset xml
+    throw std::invalid_argument("Cannot find any threshold in xml file, "
```
Mostah, line 396: nit: I would be more explicit on the exception here, the issue is that we only handle model files for this function, you could throw something like "The file given in not a model xml file"

src/DataGenerator.cpp
```diff
@@ -0,0 +1,499 @@
...
+    std::string modelpath = "../data/" + modelName;
+
+    bool saveSucceeded = doc.save_file(modelpath.c_str(), PUGIXML_TEXT("  "));
+    assert(saveSucceeded);
+  }
+}
+
+// void DataGenerator::loadDataset(std::string fileName) {}
+
+std::string DataGenerator::getXmlFileType(std::string fileName) {
```
Mostah: What if we give a wrongly formated document to this function? Can we handle the case where we don't recognise either a model or a dataset and throw an exception?

src/DataGenerator.cpp
```diff
@@ -0,0 +1,499 @@
...
+
+int DataGenerator::getNumberOfAlternatives(std::string fileName) {
+  pugi::xml_document doc;
+  std::string path = "../data/" + fileName;
+  if (!doc.load_file(path.c_str()))
+    throw std::invalid_argument("Cannot open xml file, please check path");
+
+  if (DataGenerator::getXmlFileType(fileName) == "model") {
+    // if we have a model xml
+    throw std::invalid_argument(
```
Mostah, line 406: same as above

src/DataGenerator.cpp
```diff
@@ -0,0 +1,499 @@
...
+        "most likely have a xml model file");
+  } else {
+    // if we have a dataset xml
+    pugi::xml_node node_alternatives =
+        doc.child("dataset").child("alternatives");
+    std::stringstream strValue;
+    int nb_alt;
+    strValue << node_alternatives.child_value();
+    strValue >> nb_alt;
+    return nb_alt;
```
Mostah, line 417: I'm not familiar with stringstream but isn't there a more direct way to get the value projected as `int` of `node_alternatives.child_value()`

src/DataGenerator.cpp
```diff
@@ -0,0 +1,499 @@
...
+  }
+}
+
+Performance DataGenerator::getAlternativePerformance(std::string fileName,
+                                                     std::string alt_id) {
+  pugi::xml_document doc;
```
Mostah: As you need to handle the file format and check whether if it's the right type or not for a lot of methods, you could reduce a bit the amount of code dupplicated by creating a general handler which you'll call at the begining of all your methods to check if the file type given is the one handled :)

Also for each type (model and dataset), it might be nice to have a methods to check if they have the good format or if there is an error in the architecture of the xml file. It would add the security of never causing an unexpected behaviour (such as returning empty data where we expected some)

src/DataGenerator.cpp
```diff
@@ -0,0 +1,499 @@
...
+    int nb_criteria = DataGenerator::getNumberOfCriteria(path);
+    Criteria criteria = Criteria(nb_criteria, "crit");
+    std::vector<float> alt_perf;
+
+    pugi::xml_node node_dataset = doc.child("dataset");
+    for (pugi::xml_node_iterator it = node_dataset.begin();
+         it != node_dataset.end(); ++it) {
+
+      if (it->child_value() == alt_id) {
+        // std::cout << "alternative : " << it->child_value() << std::endl;
```
Mostah: You probably forgot to remove the prints ;) (same for line 298, 402, 408, 440, 446, 448, 489) if you want to keep them because you want to print some stuff sometimes, let's start adding some logs to the project, it will probably be usefull in the future to follow what our processes are doing

src/DataGenerator.cpp
```diff
@@ -0,0 +1,499 @@
...
+          // std::cout << "crit : " << al_it->child_value() << std::endl;
+          if (strcmp(al_it->child_value(), "assignment") != 0) {
+            float perf = atof(al_it->child_value());
+            alt_perf.push_back(perf);
+          }
+        }
+        break;
+      }
+    }
+    return Performance(alt_id, criteria, alt_perf);
```
Mostah: If alt_id is not found we will return an empty Performance object, but we should probably return an error as we did not succeed in finding the alternative performance !

src/DataGenerator.cpp
```diff
@@ -0,0 +1,499 @@
...
+        const char *s = it->name();
+        std::string str(s);
+        crit_ids.push_back(s);
+      }
+    }
+  }
+  return crit_ids;
+}
+
+Criterion DataGenerator::getCriterion(std::string fileName,
```
Mostah, line 526: We probably need to handle the case of not finding crit_id here as well

src/DataGenerator.cpp
```diff
@@ -0,0 +1,499 @@
...
+          }
+        }
+        break;
+      }
+    }
+  }
+  return Criterion(crit_id, weight);
+}
+
+int DataGenerator::getAlternativeAssignment(std::string fileName,
```
Mostah, line 568: We probably need to handle the case of not finding alt_id here as well

src/DataGenerator.cpp
```diff
@@ -0,0 +1,499 @@
...
+          }
+        }
+      }
+    }
+  }
+  return 0;
+}
+
+std::vector<float>
+DataGenerator::getCriterionCategoryLimits(std::string fileName,
```
Mostah, line 608: We probably need to handle the case of not finding crit_id here as well

src/DataGenerator.h
```diff
@@ -0,0 +1,149 @@
...
+  // void saveDataset(new type, std::string datasetName);
+
+  /**
+   * Gets the xml file type
+   * returns either "model" for model xml file types or "dataset" for data xml
+   * files.
+   * @param fileName filename
+   *
+   */
+  std::string getXmlFileType(std::string fileName);
```
Mostah, line 104: Can you add the @return for the doc as well? (Same for the methods below)

test/TestDataGenerator.cpp
```diff
@@ -0,0 +1,233 @@
...
+// }
+
+// TEST(TestDataGenerator, TestModelGenerator) {
+//   DataGenerator data = DataGenerator();
+//   data.modelGenerator(2, 3, "test_model.xml");
+// }
+
+TEST(TestDataGenerator, TestXmlFileType) {
+  DataGenerator data = DataGenerator();
+  std::string type = data.getXmlFileType("test_model.xml");
```
Mostah, line 73: I suppose your test files are saved locally, it's not a problem to include them in the repo as mock data ! Otherwise the pipeline will fail anyway

Mostah (changes requested):
Great work ! Just a few fix before the merge :+1:

src/Criterion.cpp
thibmonsel: what do you mean ?

src/Criterion.cpp
thibmonsel: indeed I modified test

src/Criterion.cpp
thibmonsel: same question as above

src/Criterion.cpp
thibmonsel: I modified Criterion and Criteria Test

src/Criterion.cpp
Mostah: With a constructor using optional arguments such as 
``` 
Criterion::Criterion(std::string id, int direction = 1, float weight = 0.) {
  id_ = id;
  direction_ = direction;
  weight_ = weight;
}
```
you will be able to do `Criterion("crit0", -1, 0.6)`, `Criterion("crit1")`, `Criterion("crit2", weight=0.4)`, which are behaviours defined in the other constructors.

src/DataGenerator.cpp
thibmonsel: Added an overwrite bool that takes into account if file exists or not and if we want to overwrite it. 
Created associated test with it

src/Criterion.cpp
thibmonsel: ok I get it thanks

src/DataGenerator.cpp
thibmonsel: done added test associated with it

src/DataGenerator.cpp
thibmonsel: Done as well, created new tests

test/TestDataGenerator.cpp
thibmonsel: Remove from gitignore data directory.

src/DataGenerator.cpp
thibmonsel: This is just an arbitrary name we give in the xml file for us to read but I did do a catch statement for file names.

src/DataGenerator.cpp
thibmonsel: added some more comments. Essentially lambda is between 0.5 and 1 and I created a random number generator there that has 3 decimals

src/DataGenerator.cpp
thibmonsel: Thanks for pointing that out

src/DataGenerator.cpp
thibmonsel: ill throw an exception in there

src/DataGenerator.cpp
thibmonsel: Exception in getXmlFileType should throw an exception so we should be fine

src/DataGenerator.cpp
thibmonsel: unfortunately no we `node_alternatives.child_value ` have a type of const * char so that the way I found on stack overflow

src/DataGenerator.cpp
thibmonsel: we could definitely do that later in the project good idea

src/DataGenerator.cpp
thibmonsel: Done too

src/DataGenerator.cpp
thibmonsel: Thanks for pointing that out !

src/DataGenerator.cpp
thibmonsel: Lets do this in a second iteration

Dockerfile
```diff
@@ -4,17 +4,18 @@ FROM gcc:9.3
 # install cmake
 RUN apt-get update && apt-get -y install cmake git
 
-RUN git clone https://github.com/Mostah/fastPL.git
+RUN git clone --single-branch --branch thibault/DatasetGenerator https://github.com/Mostah/fastPL.git
```
Mostah: You might have forgotten to remove this!

Dockerfile
```diff
@@ -4,17 +4,18 @@ FROM gcc:9.3
...
 # install & configure googletest 
 RUN cd fastPL && git submodule init && git submodule update
-RUN git clone https://github.com/google/googletest.git -b release-1.10.0 
-RUN cd googletest && mkdir build && cd build && cmake .. && make && make install
-
+#RUN git clone https://github.com/google/googletest.git -b release-1.10.0
```
Mostah: same for the commented lines here

data/test.txt
```diff
@@ -0,0 +1,2 @@
+Please writr this text to a file.
```
Mostah: Not sure this file should be here as well !

src/Criteria.h
```diff
@@ -45,7 +45,7 @@ class Criteria {
    *
    * @return criterion_vect_
    */
-  std::vector<Criterion> getCriterionVect() const;
+  std::vector<Criterion> getCriterionVect();
```
Mostah: Why removing the const here?

src/Criterion.cpp
```diff
@@ -53,19 +40,24 @@ float Criterion::getWeight() const { return weight_; }
 
 void Criterion::setWeight(float weight) { weight_ = weight; }
 
+void Criterion::getRandomCriterionWeight(bool changeSeed) {
```
Mostah, line 43: I think the name of the function should be setRandomCriterionWeight. What is the purpose of having two different seed to use here though?

src/Criterion.h
```diff
@@ -37,26 +29,6 @@ class Criterion : public AtomicMCDAObject {
...
 
-  /**
-   * generateDirection generator of random direction
-   * can also be used as reinitialisation for direction
-   *
-   * @param seed Optional: seed for random generation. Set to time(NULL) by
-   * default
-   *
-   */
-  void generateDirection(unsigned long int seed = time(NULL));
```
Mostah, line 48: we won't need to use the random direction anymore?

src/DataGenerator.h
```diff
@@ -0,0 +1,188 @@
...
+
+  /**
+   * Get data from xml file
+   *
+   * @param fileName filename
+   *
+   * @return a type for model usage
+   *
+   */
+  // void loadDataset(std::string fileName);
```
Mostah, line 61: Are those 2 functions something you will implement later or where they left here by mistake?

src/DataGenerator.cpp
```diff
@@ -0,0 +1,648 @@
...
+                                "not to overwrite it");
+  }
+
+  bool saveSucceeded = doc.save_file(modelpath.c_str(), PUGIXML_TEXT("  "));
+  if (!saveSucceeded) {
+    throw std::invalid_argument("Cannot save xml file...");
+  }
+}
+
+// void DataGenerator::loadDataset(std::string fileName) {}
```
Mostah, line 194: Is this something you will add in a other PR?

test/TestCriterion.cpp
```diff
@@ -9,88 +9,42 @@ TEST(TestCriterion, TestBaseConstructorWithoutSpec) {
...
-  EXPECT_EQ("test_set", crit.getId());
-  EXPECT_EQ("criterion_test_set", crit.getName());
-  EXPECT_EQ(1, crit.getDirection());
-  EXPECT_FLOAT_EQ(0.3, crit.getWeight());
-}
-
-TEST(TestCriterion, TestDirectionGenerator) {
-  std::string id = "test_direction";
+/**
+TEST(TestCriterion, TestRandomWeightGeneration) {
```
Mostah, line 39: I thinkg we could remove this test :)

test/TestDataGenerator.cpp
```diff
@@ -0,0 +1,444 @@
+#include "../src/DataGenerator.h"
+#include "../src/utils.h"
+#include "gtest/gtest.h"
+#include <fstream>
+#include <iostream>
+#include <sstream>
+#include <tuple>
+#include <utility>
+
+TEST(TestDataGenerator, TestDummy) {
```
Mostah: This test was used for debug right? Not sure we want to add it in the codebase

test/TestDataGenerator.cpp
```diff
@@ -0,0 +1,444 @@
...
+      << "Please writr this text to a file.\n this text is written using C++\n";
+  file.close();
+  bool u = fileExists("../data/test.txt");
+  std::ostringstream os2;
+  os2 << u;
+  EXPECT_EQ(os2.str(), "1");
+}
+
+/**
+TEST(TestDataGenerator, TestDatasetGenerator) {
```
Mostah, line 11: Let's just remove or keep test, let's not add commented tests in the codebase

test/TestDataGenerator.cpp
```diff
@@ -0,0 +1,444 @@
...
+TEST(TestDataGenerator, TestDatasetGeneratorOverwrite) {
+  DataGenerator data = DataGenerator();
+  data.datasetGenerator(3, 20, 4, "test.xml", 1, 0);
+  bool u = fileExists("../data/test.xml");
+  std::ostringstream os2;
+  os2 << u;
+  EXPECT_EQ(os2.str(), "1");
+}
+
+/**
```
Mostah, line 45: same here

test/TestDataGenerator.cpp
```diff
@@ -0,0 +1,444 @@
...
+    EXPECT_EQ(err.what(),
+              std::string(" Number of criteria and the number and the length "
+                          "of the performance of the fictive alternative (ie "
+                          "profile performance) does not match"));
+  } catch (...) {
+    FAIL() << "should have throw invalid_argument error.";
+  }
+}
+
+TEST(TestDataGenerator, TestSaveModel) {
```
Mostah, line 144: This is not testing anything, we should have a check somewhere. Also, let's not have random generated models populating the code base, if you want to test the save and check if the file was saved, do it under a tmp directory and delete the directory at the end of the test to make sure we are cleaning up after the mess

Mostah (approved):
Except the few comments I added, it looks ready to be merged 😁 
In a second iteration, it could be cool to add a better check for the xml files as described in one of my comment in my previous review : we should check if the file given a known architecture and could be use as such by the rest of the appplication

src/Criteria.h
thibmonsel: I needed to remove const to some random criteria generation

src/Criterion.cpp
thibmonsel: This is present since we didn't merge Emma's branch and did a random generator.
I want to have a seed controllable random generator (we need to implement it). 
Added a task in troll

thibmonsel: trello

test/TestCriterion.cpp
thibmonsel: Lets leave it commented for now until we know if we go forward with a in-house random number generator

test/TestDataGenerator.cpp
thibmonsel: This is used to generate set, id like to keep it for now until random number generator is done

test/TestDataGenerator.cpp
thibmonsel: YEs smart modified it !

thibmonsel:
I'll merge branch when load and save dataset model is done with AlternativeAssignment type

src/Criteria.h
Mostah: Not sure if we should remove const from the getter, this prevent the behaviour of modifying the returned attribute and unintentionally modifying the original object. The getter should be a "read only function", not read and write

PR #24, opened 2021-01-06
Thomas/doxygen

Mostah:


doxygen/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+# Finds and loads settings from an external project.
+# More information at https://cmake.org/cmake/help/v3.18/command/find_package.html
+find_package(Doxygen
+# The dot component is used for graph generation.
+# More information at https://cmake.org/cmake/help/v3.18/module/FindDoxygen.html
+	COMPONENTS dot)
+
```
thibmonsel, line 7: is this normal ?

src/types/Criteria.cpp
```diff
@@ -1,5 +1,5 @@
-#include "Criteria.h"
-#include "Criterion.h"
+#include "../../include/types/Criteria.h"
+#include "../../include/types/Criterion.h"
 #include <algorithm>
 #include <iostream>
```
thibmonsel: should we start using static variable to make the include header more readable ? 
Something that would look like this ?
ie. path_to_type = ""../../include/types/"
--> #include path_to_type+"Criteria.h"

src/types/Criterion.cpp
```diff
@@ -0,0 +1,55 @@
...
+void Criterion::setId(std::string id) { id_ = id; }
+
+std::string Criterion::getName() const { return name_; }
+
+void Criterion::setName(std::string name) { name_ = name; }
+
+int Criterion::getDirection() const { return direction_; }
+
+void Criterion::setDirection(int direction) { direction_ = direction; }
+
```
thibmonsel: more recent version of Criterion on master maybe we should wait to merge other PR and do this one last

test/types/TestCriteria.cpp
```diff
@@ -50,9 +50,10 @@ TEST(TestCriteria, TestBaseConstructorWithNbOnly) {
...
   std::ostringstream os;
   os << criteria1;
-  EXPECT_EQ(os.str(), "Criteria(Criterion(id : crit0, name : , "
-                      "direction : +, weight : 0), Criterion(id : crit1, name : , "
-                      "direction : +, weight : 0), )");
+  EXPECT_EQ(os.str(),
+            "Criteria(Criterion(id : crit0, name : , "
+            "direction : +, weight : 0), Criterion(id : crit1, name : , "
+            "direction : +, weight : 0), )");
 }
```
thibmonsel: in Criterion, took out the argument name since it was useless

thibmonsel: Normally change are reflected in master or in DataGenerator branch

test/types/TestCriterion.cpp
```diff
@@ -8,7 +8,8 @@ TEST(TestCriterion, TestBaseConstructorWithoutSpec) {
   Criterion crit = Criterion(id);
   std::ostringstream os;
   os << crit;
-  EXPECT_EQ(os.str(), "Criterion(id : test_no_spec, name : , direction : +, weight : 0)");
+  EXPECT_EQ(os.str(),
+            "Criterion(id : test_no_spec, name : , direction : +, weight : 0)");
 }
```
thibmonsel: same

thibmonsel (approved):

src/types/Criterion.cpp
Mostah: Yeah no worries about this, all the modification in Criteria and Criterion was done by my linter, I did not intend to change anything here

src/types/Criteria.cpp
Mostah: I don't know how I feel about it, the path aren't that complicated for the moment and it should not really get much worse than that and it stays more readable! But if we get to a point where we have looooooong import path that we often use, let's do that 👍

doxygen/CMakeLists.txt
Mostah: Mhh yeah I guess, it's a copy paste from doxygen doc I recall. Comments can be anywhere it seems ahah

PR #28, opened 2021-01-07
Emma/alternative perf table

edix9:
**PerformanceTable:**
- slightly changed the << operator (trashing the PR, sorry)
- added a findAlt function to check is an alternative is in the table (returns a boolean) + associated test
- changed attributes to protected
- corrected a reference in a function

 **AlternativesPerformance:** 
This is a child class of PerformanceTable which has a map alternative_assignment of the assignment of category to alternatives.
- 4 constructors (2 PerformanceTable overwrited + CopyConstructor overwrited + one with PerformanceTable as input)
- getters and setters for alternative assignment map
- getters and setters for one alternative assignment
- Associated tests 

**Criterion:**
Only format


src/AlternativesPerformance.h
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * setAlternativesPerformanceMap getter of the alternatives assignments
+   *
+   * @param alt_assignment assignment map
+   */
+  void
+  setAlternativesAssignments(std::map<std::string, std::string> alt_assignment);
+
+  /**
+   * getAlternative getter of the assignment of one specified alternative
```
thibmonsel: Should we also have a setter of AA ( Alternatives assignments) if we have a Categories objects ?

src/AlternativesPerformance.h
```diff
@@ -0,0 +1,102 @@
...
+                          std::map<std::string, std::string> alt_assignment =
+                              std::map<std::string, std::string>{{"", ""}});
+
+  /**
+   * AlternativesPerformance constructor without perf values but set of
+   * criteria to evaluate performance over (PerformanceTable surcharged)
+   *
+   * @param nb_of_perfs Number of performance
+   * @param crits Criteria to evaluate performance over
+   * @param prefix Prefix to use for the name of each Performance created.
```
thibmonsel: I think we should maybe use the Category, Categories object here for the map function ?
If we keep it like this, we will never use those types and maybe make the code less readable ?
Open to discuss

src/PerformanceTable.cpp
```diff
@@ -304,4 +306,13 @@ std::vector<Perf> PerformanceTable::getWorstPerfByCrit(Criteria crits) {
...
   return worst_pv;
+}
+
+bool PerformanceTable::findAlt(std::string altName) {
+  for (std::vector<Perf> p : pt_) {
+    if (p[0].getName() == altName) {
+      return (true);
+    }
+  }
+  return (false);
```
thibmonsel: should we change true and false with 1 and 0 since that's how bool are in cpp ?

src/AlternativesPerformance.h
thibmonsel: I think it would be none sense to talk about AA without specifying the category possibilities.
By integrating that, the meta heuristic could be easier to implement ?

src/AlternativesPerformance.cpp
```diff
@@ -0,0 +1,142 @@
...
+    : PerformanceTable(perf_vect) {
+  if (mode_ != "alt") {
+    throw std::domain_error(
+        "Performance table mode should be alt to assign default categories.");
+  }
+  if (alt_assignment == std::map<std::string, std::string>{{"", ""}}) {
+    for (std::vector<Perf> pv : pt_) {
+      std::string altName = pv[0].getName();
+      alt_assignment_[altName] = "";
+    }
```
thibmonsel: If the default alt_assignment is an empty hash map, do we plan on leaving it that way or do we want to assign it random categories ? idk

thibmonsel: This discussion would have consequences on other constructors.

src/AlternativesPerformance.cpp
```diff
@@ -0,0 +1,142 @@
...
+AlternativesPerformance::getAlternativeAssignment(std::string altName) const {
+  return alt_assignment_.find(altName)->second;
+}
+
+void AlternativesPerformance::setAlternativeAssignment(std::string altName,
+                                                       std::string catName) {
+  if (this->findAlt(altName) == true) {
+    alt_assignment_[altName] = catName;
+  } else {
+    throw std::invalid_argument("The alternatives in the map should be present "
```
thibmonsel: You can change (this->findAlt(altName) == true) to (this->findAlt(altName)) if you'd like less code is better :)

src/AlternativesPerformance.h
```diff
@@ -0,0 +1,102 @@
...
+   *
+   * @param altName name of the alternative the category will be assigned to
+   * @param catName name of the category to assign
+   */
+  void setAlternativeAssignment(std::string altName, std::string catName);
+
+private:
+  // Hashmap: key = Alternative Name, value = Category Name
+  std::map<std::string, std::string> alt_assignment_;
+};
```
thibmonsel: if we want to integrate Category, Categories type this might need to change.

src/AlternativesPerformance.h
```diff
@@ -0,0 +1,102 @@
...
+  AlternativesPerformance(int nb_of_perfs, Criteria &crits,
+                          std::string prefix = "alt",
+                          std::map<std::string, std::string> alt_assignment =
+                              std::map<std::string, std::string>{{"", ""}});
+
+  /**
+   * AlternativesPerformance constructor using an existing performance table
+   *
+   * @param alt_assignment Map of alternative assignements to categories
+   * @param perf_table Performance table to copy
```
thibmonsel: maybe have a constructor that gives as input an vector of alternative ids could be useful !
AA = AlternativeAssignment
ie AA(std::vector<std::string> ['a1,'a2'], int nbCategories) or/and AA(std::vector<std::string> ['a1,'a2'], Categories cat)

src/AlternativesPerformance.cpp
```diff
@@ -0,0 +1,142 @@
...
+  out << "], AlternativesAssignment{ ";
+
+  auto it = alt.alt_assignment_.begin();
+  while (it != alt.alt_assignment_.end()) {
+    out << it->first << "->" << it->second << " ";
+    it++;
+  }
+  out << "}";
+  return out;
+}
```
thibmonsel, line 110: Do we need to view the performance table when we are dealing with the AA type ?

test/TestAlternativesPerformance.cpp
```diff
@@ -0,0 +1,268 @@
...
+  std::ostringstream os2;
+  os2 << alt_perf2;
+  EXPECT_EQ(
+      os2.str(),
+      "AlternativesPerformance( PerformanceTable[ Performance: Perf( name : "
+      "test0, crit : a0, value : 0 ) Perf( name : test0, crit : a1, value : 0 "
+      ") | Performance: Perf( name : test1, crit : a0, value : 0 ) Perf( name "
+      ": test1, crit : a1, value : 0 ) | ], AlternativesAssignment{ test0->cat0"
+      " test1->cat1 }");
+  EXPECT_EQ(alt_perf2.getMode(), "alt");
```
thibmonsel: Names tend to confuse : name of a performance should be 'a1' (alternative1) or 'b1' (profile1) if we consider notations from MDCA.
Criteria could be : crit1, etc....

thibmonsel (changes requested):
Put some comments :) We should have a discussion on those comments to validate the approach. Might as well take extra time on type elaboration than going back to change it when we will implement the Algorithm !

src/AlternativesPerformance.cpp
Mostah: In a sense it's useless to use an AP object if we don't use categories as it would just be a performance table. But does it make sense to have randomly assigned categories? I would need to think that through a bit more but given the thesis we can't really assign categories randomly and stay consistent right?

src/AlternativesPerformance.cpp
Mostah: Similarily you have a few `if( condition == false)`, you could simplify with just using `if(!condition)`

src/Criterion.cpp
```diff
@@ -2,26 +2,26 @@
 #include <iostream>
```
Mostah: fyi we all have these changes in our PR, it is likely to cause annoying merge conflict after the first will be merged !

src/PerformanceTable.h
```diff
@@ -142,7 +143,16 @@ class PerformanceTable {
...
 
-private:
+  /**
+   * findAlt return true if the alternative is in the performance table
+   *
+   * @param altName Alternative to lookup
+   *
+   * @return true if found, false if nots
+   */
+  bool findAlt(std::string altName);
```
Mostah: nit: by convention we tend to use a question that can be answered by yes or no for the name of boolean functions: 

example: `isAltInMap(std::string altName)`

it makes the code more readable when we read the code in conditions: 

```
if ( isAltInMap("alt1") ) { }
```

test/TestAlternativesPerformance.cpp
Mostah: Not sure we followed this notation for the previous tests, but indeed it will be more understandable if we start using it ! Especially if we turn our tests in documentation later :)

src/AlternativesPerformance.h
edix9: We can do a setter of AA with a Cateogory but I don't see how we can do with Categories how do you manage the assignment?)

src/AlternativesPerformance.h
edix9: Do we really want to have Category objects in the hash map? I thought the hash map was just for the ids and from that we could deduce the Category. 
If we change to category then we would also have to change the alternative id to a Performance object to be consistent.

src/AlternativesPerformance.cpp
edix9: Maybe the AP object can be created with blank categories in the main model function and the categories can be assigned afterwards, but sure this is kind of useless to have it like that. Anyway the default map value might never be used, not sure this means we should remove it

src/AlternativesPerformance.cpp
edix9: will do thanks!

src/AlternativesPerformance.h
edix9: Why use a vector and not use directly the hashmap: AA(std::map<std::string, std::string>{{"a1", ""},{"a2", ""}}?
About the category/categories object we should definitely discuss that!

src/AlternativesPerformance.cpp
edix9: Might be useful to check the results?
But we could do another function like "printAlternativesAssignments" to show only the AA

test/TestAlternativesPerformance.cpp
edix9: I used the same notation as Thomas to be consistent lol :) but right this is confusing sometimes, I'll correct it

src/PerformanceTable.h
edix9: ok, will correct!

src/AlternativesPerformance.cpp
Mostah: It's fine keeping it that way, maybe in the doc we could specify that we should not use an AP with the default empty categories and use a pt instead!

edix9: Ok I'll add that

src/AlternativesPerformance.h
```diff
@@ -1,28 +1,34 @@
...
 #include "PerformanceTable.h"
 #include <iostream>
 #include <iterator>
-#include <map>
 #include <ostream>
 #include <string.h>
+#include <unordered_map>
 #include <vector>
 
+extern Category default_cat;
```
Mostah, line 15: nit: I think it's a good practice to put default value as const and use a naming all in caps letters (DEFAULT_MAP)

thibmonsel (approved):

PR #29, opened 2021-01-13
Thomas/mr sort model

Mostah:


src/MRSortModel.h
```diff
@@ -0,0 +1,72 @@
...
+  /**
+   * categoryAssignment assign the categories given the performance table
+   * and the current state of the model.
+   *
+   * @param pt PerformanceTable
+   *
+   * @return category_assignment pair vector with first value is the alternative
+   * id and second value the category assigned.
+   */
+  std::vector<std::pair<std::string, std::string>>
```
Mostah: It could makes sense here to have an AlternativePerformance return, however we miss the link to the categories from the model perspective. If it's worth it, I could look into it but otherwise it might overcomplicate some parts

src/MRSortModel.cpp
```diff
@@ -0,0 +1,77 @@
...
+#include <map>
+#include <string>
+
+MRSortModel::MRSortModel(Criteria &crits, PerformanceTable &profs, float lbd,
+                         std::string id)
+    : criteria(crits), profiles(profs) {
+  if (!profiles.isOrdered()) {
+    throw std::invalid_argument(
+        "The given performance table cannot be use as a profile");
+  }
```
thibmonsel: Match the name of the PerformanceTable argument with the header file ie PerformanceTable &profiles.

Mostah: Yeah this was done in purpose, I like that fact that in the header file and thus in the doc we have explicit variable names, however if I keep them that way in the cpp file the implementation is a bit disturbing as they would have the same name as the class attributes: 
`criteria(crits), profiles(profs)` would become `criteria(criteria), profiles(profiles)`

open to suggestion though I was out of idea so I found this turnaround

include/types/MRSortModel.h
```diff
@@ -0,0 +1,72 @@
...
+   * and the current state of the model.
+   *
+   * @param pt PerformanceTable
+   *
+   * @return category_assignment pair vector with first value is the alternative
+   * id and second value the category assigned.
+   */
+  std::vector<std::pair<std::string, std::string>>
+  categoryAssignment(PerformanceTable &pt);
+
```
thibmonsel: We should probably use the type that we created. If will be easier to debug and we will have better code readability.
Either we should return an unordered_map (same as in AlternativePerformance) or and AlternativePerformance type.

thibmonsel: We should also make a function that takes as input one Performance type and send its assigned Category type

src/types/MRSortModel.cpp
```diff
@@ -0,0 +1,77 @@
...
+  for (std::vector<Perf> alt : pt.getPerformanceTable()) {
+    bool assigned = false;
+    for (std::vector<Perf> profile_h : profiles.getPerformanceTable()) {
+      float c = 0;
+      for (Perf perf_j : alt) {
+        for (Perf prof_i : profile_h) {
+          if ((prof_i.getCrit() == perf_j.getCrit()) &&
+              (perf_j.getValue() > prof_i.getValue())) {
+            c = c + criteria[perf_j.getCrit()].getWeight();
+          }
```
thibmonsel: should be good to put in documentation how you are going through the profiles in ascending or descending order

thibmonsel: For category assignment you must check that for each the weights for an alternative a is higher for class h but lower to class h+1 to assign correctly the category I don't see that here.

test/types/TestMRSortModel.cpp
```diff
@@ -0,0 +1,78 @@
...
+  std::vector<Performance> perf_vect;
+  Criteria crit = getTestCriteria();
+  std::vector<float> given_perf0 = {0.8, 1, 0.4};
+  std::vector<float> given_perf1 = {0.8, 0.1, 0.3};
+  std::vector<float> given_perf2 = {0.6, 0, 0.2};
+  perf_vect.push_back(Performance(crit, given_perf0, "cat0"));
+  perf_vect.push_back(Performance(crit, given_perf1, "cat1"));
+  perf_vect.push_back(Performance(crit, given_perf2, "cat2"));
+  return PerformanceTable(perf_vect);
+}
```
thibmonsel: Do those functions belong in the test file ?

thibmonsel (changes requested):
Need to review some comments in order that we are all on the same page

include/types/MRSortModel.h
Mostah: Yes I agree, it would be better here to return an AlternativesPerformance, however I was kinda block on the matter: from a model perspective, I don't have access to the Category object in the current implementation. 
I will add the Categories object of the problem to the definition of the model (by reference) to fill the gap

Mostah: I could do the one with Performance -> Category if you think we will have use of it, but as a Performance table is not a vector<Performance> anymore, we won't be able to iterate over a PerformanceTable and use this new function.
 Let me know if you still think it will be useful to add it for the future

src/MRSortModel.h
thibmonsel: Wont the number of categories be a "hyper parameter" when training ? 
If so we can easily create an AlternativePerformance

include/types/MRSortModel.h
thibmonsel: You can also access the category via the nb of row of your profile PerformanceTable !
If you got 4 profiles that means that you have 5 classes so you can reconstruct the Category object and not pass it into parameter. 👍

thibmonsel (approved):

src/types/MRSortModel.cpp
Mostah: You are right it's a bit shady, I will add some more comments, fyi the profiles are stored in descending order (the highest being the first row of the profile pt), thus by itterating over the profile and assigning as soon as it's possible, we guarantee that the sum of weight for an alterntive is indeed higher for class h and lower for class h+1 (it would have been set to h+1 otherwise). 

About that matter I still think it would make sense to have a Profile class that set a common assumption about the order of the profile and guarantee its integrity. I will probably add that in this PR a bit later, it will make things more clear for everyone

test/types/TestMRSortModel.cpp
Mostah: Yes they just generate some data, it makes my tests a bit more readable and can be reused for further tests as well :)

include/types/MRSortModel.h
Mostah: By doing that we take the risk to have different names between the initial categories and those ones as the initial ones could have been created with a prefix we don't know of... This might be a particular situation but it can be a breaking point so we will probably need to pass the Categories to model to be safe

src/types/Profiles.cpp
```diff
@@ -0,0 +1,71 @@
...
+    }
+    std::sort(r_vect.begin(), r_vect.end());
+    std::reverse(r_vect.begin(), r_vect.end());
+    for (int i = 0; i < pt_.size(); i++) {
+      pt_[i][j].setValue(r_vect[i]);
+    }
+  }
+}
+
+bool Profiles::isProfileOrdered() {
```
Mostah, line 61: We do not guarantee that we do not have equal profiles (equal on each criterion) here, which could be an issue for later

include/types/MRSortModel.h
```diff
@@ -0,0 +1,74 @@
...
+class MRSortModel {
+public:
+  /**
+   * MRSortModel standard constructor
+   *
+   * @param criteria Criteria object
+   * @param profiles PerformanceTable containing the profiles
+   * @param categories Categories object used for the profiles
+   * @param lambda threshold
```
thibmonsel: Change comment for PT to Profiles

src/types/MRSortModel.cpp
```diff
@@ -0,0 +1,99 @@
...
+}
+
+MRSortModel::MRSortModel(int n_cat, int n_crit, std::string id)
+    : criteria(n_crit), profiles(n_cat - 1, criteria, "prof"),
+      categories(default_cats) {
+  if (n_cat < 2) {
+    throw std::invalid_argument(
+        "The number of categories (n_cat) must be greater than 2");
+  }
+  id_ = id;
```
thibmonsel: greater than 1

thibmonsel (approved):

PR #32, opened 2021-01-14
Added Save and Load Dataset methods to finish clss

thibmonsel:
Add 2 new functions and tests with it saveDataset() and loadDataset().

For loadDataset() : 
From the xml file we get the data and return an perftable and unordered_map from it. 

For saveDataset():
With AlternativesPerformance, PerformanceTable and the number of categories we save the content in an xml file

### DIFF
--- a/data/test.xml
+++ b/data/test.xml
@@ -4,7 +4,7 @@
   <datasetName>test.xml</datasetName>
   <criteria>3</criteria>
   <categories>4</categories>
-  <alternatives>20</alternatives>
+  <alternatives>5</alternatives>
   <alternative>alt0<crit0>0.000000</crit0>
     <crit1>0.000000</crit1>
     <crit2>0.000000</crit2>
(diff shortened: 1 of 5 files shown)


data/test.xml
```diff
@@ -4,7 +4,7 @@
   <datasetName>test.xml</datasetName>
```
Mostah: Can you rename this file to something related to dataset? So we know that this on is used in dataset unit test

src/types/DataGenerator.cpp
```diff
@@ -191,44 +192,51 @@ void DataGenerator::modelGenerator(int nb_criteria, int nb_categories,
...
 std::tuple<float, Criteria, PerformanceTable>
 DataGenerator::loadModel(std::string fileName) {
+  pugi::xml_document doc = DataGenerator::openXmlFile(fileName);
+
+  if (DataGenerator::getXmlFileType(fileName) != "model") {
+    // if we have a model xml
+    throw std::invalid_argument(
+        "Cannot find any lambda and profile performances in xml file, "
+        "most likely have a dataset xml file");
+  } else {
```
Mostah, line 204: nit: the else statement is not necessary here as if you get into the if statement the run of the function will stop anyway :)

src/types/DataGenerator.cpp
```diff
@@ -319,6 +328,146 @@ void DataGenerator::saveModel(std::string fileName, float lambda,
   }
 }
 
+std::tuple<PerformanceTable, std::unordered_map<std::string, Category>>
+DataGenerator::loadDataset(std::string fileName) {
```
Mostah: Should we return an AlternativesPerformance here instead of the tuple ?

Mostah (approved):
It looks good to me ! Just wondering, should we keep the commented test? Would you like to make them work in the future?

data/test.xml
thibmonsel: done

src/types/DataGenerator.cpp
thibmonsel: took it out

src/types/DataGenerator.cpp
thibmonsel: yup changed

test/types/TestDataGenerator.cpp
```diff
@@ -7,21 +7,19 @@
...
   EXPECT_EQ(os2.str(), "1");
 }
-*/
 
 TEST(TestDataGenerator, TestDatasetGeneratorNotOverwrite) {
   DataGenerator data = DataGenerator();
   try {
-    data.datasetGenerator(3, 20, 4, "test.xml", 0, 0);
+    data.datasetGenerator(3, 5, 4, "test_dataset.xml", 0, 0);
     FAIL() << "should have throw invalid_argument error.";
```
edix9, line 23: not important but throw**n**

PR #33, opened 2021-01-14
Thomas/run app

Mostah:


thibmonsel (approved):
Looks good, I don't have much experience in app running but nothing stands out of the ordinary

src/app.cpp
```diff
@@ -0,0 +1,76 @@
...
+#include <iostream>
+
+void App::initializeLogger(YAML::Node &yml_conf) {
+  std::string log_path;
+  if (yml_conf["log_file"]) {
+    log_path = yml_conf["log_file"].as<std::string>();
+  } else {
+    log_path = "../logs/app_log.txt";
+  }
+  conf.logger = spdlog::basic_logger_mt("app_logger", "../logs/app_log.txt");
```
Mostah: nit: should be log_path here

src/main.cpp
```diff
@@ -1,17 +1,33 @@
...
-  spdlog::set_level(spdlog::level::info);
-}
+#include "../include/app.h"
+
+int main(int argc, char *argv[]) {
+  YAML::Node yaml_config = YAML::LoadFile("../app-config.yaml");
+  App app = App();
+
+  int exec_code = app.parseArgs(argc, argv);
+  if (exec_code == 1) {
```
Mostah:
```
if( exec_code != -1) {
   return exec_code;
}

File: src/main.cpp [hunk 3]
```diff
@@ -1,17 +1,33 @@
...
 #include <iostream>
 
-void initialize_logger() {
-  auto logger = spdlog::basic_logger_mt("app_logger", "../logs/app_log.txt");
-  spdlog::set_level(spdlog::level::info);
-}
+#include "../include/app.h"
+
+int main(int argc, char *argv[]) {
+  YAML::Node yaml_config = YAML::LoadFile("../app-config.yaml");
```
Mostah: Need to Catch exception in case file not found

app-config.yaml
```diff
@@ -0,0 +1,10 @@
+env:
+  staging:
+    log_level: "DEBUG"
+
+  prod:
+    log_level: "INFO"
+# log_level takes its value in { DEBUG, INFO, WARN, ERROR, CRITICAL }
+
+log_file: ../logs/app_log.txt
+data_dir: ../data/
```
Mostah, line 10: Need to use this path in source files instead or harcoded one

thibmonsel (approved):

PR #37, opened 2021-01-20
Emma/random uniform numbers

edix9:
added random int and float generation functions


src/types/Criterion.cpp
```diff
@@ -41,14 +27,19 @@ float Criterion::getWeight() const { return weight_; }
 
 void Criterion::setWeight(float weight) { weight_ = weight; }
 
-void Criterion::getRandomCriterionWeight(bool changeSeed) {
-  if (changeSeed) {
-    Criterion::setWeight(getRandomUniformNumber(1));
+void Criterion::generateDirection(unsigned long int seed) {
+  srand(seed);
+  if (((float)rand() / RAND_MAX) < 0.5) {
```
Mostah: I think you could use your getRandomUniformFloat here :)

Mostah (approved):

PR #39, opened 2021-01-20
Thibault/initialize profile

thibmonsel:
Profiles class : 
- Profile object is a ordered PT table in crit mode. 
- Example is below
 Performance([Performance(Perf(name: crit0, crit: cat0, value:0,1), Perf(name: crit0, crit: cat1, value:0,2), Perf(name: crit0, crit: cat2, value:0,3)), Performance(Perf(name: crit1, crit: cat0, value:0,2), Perf(name: crit0, crit: cat1, value:0.3), Perf(name: crit0, crit: cat2, value:0,4)]
- We also have a display function that show the profiles in the "alt" mode in a nice way even though the Profiles objects only have one mode "crit" and are represented in that mode. ("alt" does not exist for Profiles).

InitializeProfile class: 
- that initializes the profiles
- Uses the MRsort model and changes its PT of profiles.

Changed the random number generator to have truly random number generated vectors.

Regarding the tests.
Not able to initialise properly the profile of a MRsort model with the method generateRandomPerfValues of PT Class. This is normal since the random generator doesn't really comply with our initial hypothesis of ordered categories. (With that random generator the Profile constructor fails at the if statement that verifies the orderliness of the profile.

I might have missed some other details but InitProf looks done.

Discussion :  For displaying in a nice way Profiles, do we change the atomic block of std::vector Perf that requires to have all of its name identical.


include/utils.h
```diff
@@ -112,5 +114,20 @@ randomCategoriesLimits(int nbCategories, unsigned long int seed = time(NULL)) {
...
+ * Overloading << operator for std::unordered_map
+ *
+ * @param out ostream
+ * @param std::unordered_map  object
+ *
+ */
+template <typename K, typename V>
+std::ostream &operator<<(std::ostream &out, std::unordered_map<K, V> const &m) {
+  for (auto const &pair : m) {
+    out << "{, (" << pair.first << ": " << pair.second << "),";
```
edix9, line 131: Either we need to remove this if we don't feel like using it, or we need to adapt the << operator in alternativePerformance so that we use it and it matches in the tests

test/TestMain.cpp
```diff
@@ -1,14 +1,15 @@
...
-#include "types/TestCriteria.cpp"
-#include "types/TestCriterion.cpp"
-#include "types/TestDataGenerator.cpp"
-#include "types/TestMRSortModel.cpp"
-#include "types/TestPerf.cpp"
-#include "types/TestPerformance.cpp"
-#include "types/TestPerformanceTable.cpp"
+#include "learning/TestInitializeProfile.cpp"
+// #include "TestUtils.cpp"
+// #include "types/TestAlternativesPerformance.cpp"
```
edix9: need to uncomment this

include/learning/InitializeProfile.h
```diff
@@ -0,0 +1,126 @@
+#ifndef InitializeProfile_H
```
Mostah: Can we change these variable names as well as the name of the files to be the same as the class name?

include/learning/InitializeProfile.h
```diff
@@ -0,0 +1,126 @@
...
+   * @param config config setup for logging
+   * @return ProfileInitializer object
+   */
+  ProfileInitializer(Config &config, const ProfileInitializer &profInit);
+
+  /**
+   * ProfileInitializer standard deconstructor
+   *
+   */
+  ~ProfileInitializer(){};
```
Mostah: nit: the implementation of the destructor should be in the cpp file:

in the .h file:
```
~ProfileInitializer();
```

in the .cpp file:
```
ProfileInitializer::~ProfileInitializer(){ }
```

include/learning/InitializeProfile.h
```diff
@@ -0,0 +1,126 @@
...
+   * @param delta articifial integer used to compare each candidates with the
+   * imaginary profile performanc.
+   *
+   * @return probability of choosing alt_id for crit.
+   */
+  float weightedProbability(const std::string altId, const Criterion &crit,
+                            const Category &catAbove, const Category &catBelow,
+                            const int nbCategories,
+                            const std::vector<float> &catFrequency,
+                            std::vector<std::string> candidates,
```
Mostah: candidates should be passed by reference here!

include/learning/InitializeProfile.h
```diff
@@ -0,0 +1,126 @@
...
+   * given by the metaheuristic.
+   *
+   * @param categories Categories object
+   * @param MRSortModel Mrsort model object
+   */
+  void initializeProfiles(Categories &categories, MRSortModel &model);
+
+private:
+  Config &conf;
+  AlternativesPerformance altPerformance_;
```
Mostah: As the alternative performance does not belong exclusively to the ProfileInitializer, it will be initialized previously in the pipeline. Thus, it should not be stored by value here but by reference to avoid dupplicating the object :)

include/learning/InitializeProfile.h
```diff
@@ -0,0 +1,126 @@
...
+                               const std::vector<float> &catFre);
+
+  /**
+   * Updates profile attribute from MRSortModel class with the a new profile
+   * given by the metaheuristic.
+   *
+   * @param categories Categories object
+   * @param MRSortModel Mrsort model object
+   */
+  void initializeProfiles(Categories &categories, MRSortModel &model);
```
Mostah: What are those categories for? Isn't this object the same as the one contained in the model?

include/types/Categories.h
```diff
@@ -89,6 +89,14 @@ class Categories {
    */
   Category getCategoryOfRank(int rank);
 
+  /**
+   * Set a new values of Category ids from Categories object
```
Mostah, line 93: You might want to update this doc :)

include/types/MRSortModel.h
```diff
@@ -31,6 +31,7 @@ class MRSortModel {
    *
    * @param ap alternative performance object
    * @param id optional name of the model
+   * @param mode give the mode of the profile performance table
    */
```
Mostah, line 35: Looks like you should remove this changes as well. Btw can you update this doc to be the right one for this constructor? Looks like a copy paste mistake from my previous PR !

include/types/Profiles.h
```diff
@@ -38,14 +40,21 @@ class Profiles : public PerformanceTable {
    */
   Profiles(const Profiles &profiles);
 
-  ~Profiles();
+  ~Profiles(){};
```
Mostah, line 47: Same comment as for the initialize profile, shouldn't be the implementation only in cpp file?

src/learning/InitializeProfile.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+  std::transform(map.begin(), map.end(), values.begin(), cat_value_selector);
+
+  // Counting the number of occurence for each category rank
+  std::unordered_map<int, int> counts;
+  for (auto v : values)
+    ++counts[v];
+
+  // Counting the number of alternatives in the altPerformance object
+  const int nbAlternatives = std::accumulate(
```
Mostah: Wouldn't .size() in the performance table be a easier way to get nbAlternatives?

src/learning/InitializeProfile.cpp
```diff
@@ -0,0 +1,203 @@
...
+  // Computing the frequency
+  std::vector<float> frequency;
+  for (auto const pair : counts) {
+    float c = pair.second;
+    frequency.push_back(c / nbAlternatives);
+  }
+  return frequency;
+}
+
+std::vector<std::string> ProfileInitializer::getProfilePerformanceCandidates(
```
Mostah: In general this function could be implemented in a fastest way, the for loop for only one profile is ineficient: we could go over the alternativesAssignments only once and compute all the candidates for each profile at the same time instead of going over the aa n_profiles time

src/learning/InitializeProfile.cpp
```diff
@@ -0,0 +1,203 @@
...
+}
+
+float ProfileInitializer::weightedProbability(
+    const std::string altId, const Criterion &crit, const Category &catAbove,
+    const Category &catBelow, const int nbCategories,
+    const std::vector<float> &catFrequency, std::vector<std::string> candidates,
+    float delta) {
+  // creating imaginary profile performance for criterion crit
+  std::string critId = crit.getId();
+  float imaginaryProfilePerformance =
```
Mostah: In terms of optimization here it's not great, getPerf is in O(n_prof * n_crit). The Performance table should be itterated over directly to transform this operation in O(1). 

Consider changing the function to 
```
float ProfileInitializer::weightedProbability(
    &Perf perfs, const Category &catAbove,
    const Category &catBelow, const int nbCategories,
    const std::vector<float> &catFrequency, std::vector<std::string> candidates,
    float delta) {
```

src/learning/InitializeProfile.cpp
```diff
@@ -0,0 +1,203 @@
...
+  int catBelowCounter = 0;
+
+  // candidates are computed thanks to the method
+  // ProfileInitializer::getProfilePerformanceCandidates(crit, catBelow,
+  // nbCategories);
+  for (std::string can : candidates) {
+    // if the performance of the candidate is higher than the profile
+    // performance and that the alternatives category is catAbove then we have
+    // correctly classified it
+    if (altPerformance_.getPerf(can, critId).getValue() >
```
Mostah: getPerf here and in L130 should be avoided as well, I see you might need to give the whole row of crit to this function instead of a single perf.

src/learning/InitializeProfile.cpp
```diff
@@ -0,0 +1,203 @@
...
+    for (int i = 1; i < candidates.size(); i++) {
+      if (tmp < randomNumber) {
+        tmp += altProba[i];
+      } else {
+        index = i - 1;
+        break;
+      }
+    }
+    categoryLimits.push_back(
+        altPerformance_.getPerf(candidates[index], crit.getId()).getValue());
```
Mostah: Same here, consider not using getPerf here, you might need to add a new function in perfTable to get all the Perf you need in one shot

src/types/PerformanceTable.cpp
```diff
@@ -18,7 +22,9 @@ PerformanceTable::PerformanceTable(std::vector<Performance> &perf_vect) {
     // ensure there is no performance with dupplicated name
     if (std::find(perf_id_vect.begin(), perf_id_vect.end(), p.getId()) !=
         perf_id_vect.end()) {
-      throw std::invalid_argument("Each performance must have different ids.");
```
Mostah: What is the issue with this check here? It's an important assumption

src/types/PerformanceTable.cpp
```diff
@@ -92,13 +98,18 @@ std::vector<Perf> PerformanceTable::operator[](std::string name) const {
 void PerformanceTable::generateRandomPerfValues(unsigned long int seed,
                                                 int lower_bound,
                                                 int upper_bound) {
+  int i = 0;
   if (lower_bound > upper_bound) {
     throw std::invalid_argument(
         "Lower bound must be lower than the upper bound.");
   }
+
+  std::mt19937 gen(seed);
```
Mostah: If that's the way of getting a real uniform distribution, might be better to change getRandomUniformFloat instead !

src/types/Profiles.cpp
```diff
@@ -44,11 +52,13 @@ void Profiles::generateRandomPerfValues(unsigned long int seed, int lower_bound,
...
         "Lower bound must be lower than the upper bound.");
   }
-  srand(seed);
+  std::mt19937 gen(seed);
+  std::uniform_real_distribution<> dis(lower_bound, upper_bound);
+
   for (int j = 0; j < pt_[0].size(); j++) {
     std::vector<float> r_vect;
     for (int i = 0; i < pt_.size(); i++) {
-      r_vect.push_back(getRandomUniformFloat(seed, lower_bound, upper_bound));
```
Mostah, line 51: same comment here as above the random changes should be made in utils

Mostah (changes requested):
Nice work, we are almost there! A commented a few optimization thought, you might need to rethink your structure to make it work but most of your code is already there! 

Also, before considering merging this, we should make a test on a bigger example (maybe 2-3 crit and 6-8 cats?) to have a true example to show ?

include/learning/InitializeProfile.h
thibmonsel: Sounds good

thibmonsel: maybe add a const to make sure we don't touch the object ? 
Probably not a good idea since for each iteration we would need to change the AlternativesPerformance object ?

include/learning/InitializeProfile.h
thibmonsel: Ah nice didn't see that categories was in the MRSort model ill take out this argument

include/learning/InitializeProfile.h
Mostah: Why would we need to change the AlternativesPerformance object at each iteration? shouldn't this be constant from the begining (dataset input)?

thibmonsel: Actually problem with passing a reference here. 
For the constructor we have : 
` non-const lvalue reference to type 'AlternativesPerformance' cannot bind to a temporary of type
      'AlternativesPerformance'`

Temporaries can not be bound to non-const lvalue references.

src/learning/InitializeProfile.cpp
thibmonsel: Dont get it

src/types/PerformanceTable.cpp
thibmonsel: uncommented, brought it back to initial state

include/learning/InitializeProfile.h
Mostah: You'll find the way to properly pass this thing ! 😄

src/types/Profiles.cpp
thibmonsel: done

src/types/PerformanceTable.cpp
thibmonsel: done

include/utils.h
thibmonsel: Did overwrite the << operator to include this unordered map printing

src/learning/InitializeProfile.cpp
thibmonsel: Same here do not understand

Mostah: See two comments above: getPerf is not efficient. We will probably need to find a way to not call those and just pass the Perf objects to the functions

include/learning/InitializeProfile.h
thibmonsel: solved

src/learning/InitializeProfile.cpp
thibmonsel: still don't get it ahah

src/learning/InitializeProfile.cpp
thibmonsel: need to not use [] operator since it does a search in a n**2 matrix and is slow. Better use to go through PT once and get std::Vector of Perf

thibmonsel: done

include/learning/ProfileInitializer.h
```diff
@@ -0,0 +1,125 @@
...
+#include "../types/Profiles.h"
+
+class ProfileInitializer {
+public:
+  /**
+   * ProfileInitializer standard constructor.
+   *
+   * @param altPerfs AlternativesPerformance objet that represents our dataset
+   * @param config config setup for logging
+   * @return ProfileInitializer object
```
Mostah: The constructors aren't returning anything :)

include/learning/ProfileInitializer.h
```diff
@@ -0,0 +1,125 @@
...
+
+  /**
+   * Get ids of potential alternative candidates to initialize the profile
+   * performance for one specific criterion and category
+   *
+   * @param crit Criterion object
+   * @param cat Category object
+   * @param nbCategories number of categories in our dataset
+   *
+   * @return vector of alternative ids
```
Mostah, line 68: nit: might need to update this doc with the new return type

include/learning/ProfileInitializer.h
```diff
@@ -0,0 +1,125 @@
...
+  std::vector<Perf> getProfilePerformanceCandidates(const Criterion &crit,
+                                                    const Category &cat,
+                                                    const int nbCategories);
+
+  /**
+   * Compute the likelihood of choosing the performance of the alternative
+   * identified by its altId as the performance value for profile delimiting
+   * catAbove and catBelow for criterion crit.
+   *
+   * @param altId alternative id
```
Mostah: Same here, it needs a little update!

src/learning/ProfileInitializer.cpp
```diff
@@ -0,0 +1,214 @@
...
+
+  // Category rank starts at 0, worst category is 0
+  if (cat.getCategoryRank() == nbCategories - 1) {
+    catAbove = cat.getCategoryRank();
+    catBelow = catAbove - 1;
+  } else {
+    catBelow = cat.getCategoryRank();
+    catAbove = catBelow + 1;
+  }
+  // std::cout << altPerformance_.getPerformanceTable();
```
Mostah: We can probably remove this comment !

src/learning/ProfileInitializer.cpp
```diff
@@ -0,0 +1,214 @@
...
+std::vector<Perf> ProfileInitializer::initializeProfilePerformance(
+    const Criterion &crit, Categories &categories,
+    const std::vector<float> &catFre) {
+  std::vector<float> categoryLimits;
+  int nbCategories = categories.getNumberCategories();
+
+  for (int i = 0; i < nbCategories - 1; i++) {
+    std::vector<float> altProba;
+    std::vector<Perf> candidates =
+        ProfileInitializer::getProfilePerformanceCandidates(crit, categories[i],
```
Mostah, line 165: You might want to double check that, but I think if the method is not static, it should be call with `this->`.
However, we constructed the algorithm to have our methods independent from the state, so that could be a good idea to put most of the method in static and call them as you did here!

src/learning/ProfileInitializer.cpp
```diff
@@ -0,0 +1,214 @@
...
+  }
+  std::reverse(categoryLimits.begin(), categoryLimits.end());
+  // this is a work around since we would actually need to construct a
+  // Performance with a categories object.
+  std::vector<Perf> vect_p;
+  for (int i = 0; i < categoryLimits.size(); i++) {
+    // Cannot give a nice name to it since each vector of Perf need to have
+    // same name
+    vect_p.push_back(
+        Perf(crit.getId(), "cat" + std::to_string(i), categoryLimits[i]));
```
Mostah, line 198: I don't think we should use `"cat" + std::to_string(i)` as a name here. If the initial model was named differently, we could have an issue later if we expect the names not to change!

src/learning/ProfileInitializer.cpp
```diff
@@ -0,0 +1,214 @@
...
+  }
+  return vect_p;
+}
+
+void ProfileInitializer::initializeProfiles(MRSortModel &model) {
```
Mostah, line 203: Just a note for later: memory wise this could be optimized by changing in place the values of the profiles instead of creating a new object.

src/types/PerformanceTable.cpp
```diff
@@ -92,10 +96,12 @@ std::vector<Perf> PerformanceTable::operator[](std::string name) const {
 void PerformanceTable::generateRandomPerfValues(unsigned long int seed,
                                                 int lower_bound,
                                                 int upper_bound) {
+  int i = 0;
```
Mostah: Doesn't look like you are using this variable

src/types/Profiles.cpp
```diff
@@ -1,38 +1,46 @@
...
 
 Profiles::Profiles(std::vector<Performance> &perf_vect)
     : PerformanceTable(perf_vect) {
-  if (!this->isProfileOrdered()) {
-    throw std::invalid_argument(
-        "The given performance vector cannot be used as a profiles performance "
-        "table. Each row must be ranked such that for each row i we have on "
-        "each criterion j : v_i - 1_j > v_i_j > v_i + 1_j");
-  }
+  mode_ = "crit";
```
Mostah: It could be a great time to start using logs here: let's log a warning if the profile is not in an ordered state, so we can keep a trace of that later !

include/utils.h
```diff
@@ -99,8 +103,10 @@ inline int getRandomUniformInt(unsigned long int seed = 0, int min = 0,
  */
 inline float getRandomUniformFloat(unsigned long int seed = 0, float min = 0,
```
Mostah, line 104: I think we should use Time(0) here instead of 0 in default value of the generator ! Thus the default way of using this function is a real random generator between 0 and 1 ! (same for the function above for int)

src/types/Profiles.cpp
```diff
@@ -44,11 +52,12 @@ void Profiles::generateRandomPerfValues(unsigned long int seed, int lower_bound,
     throw std::invalid_argument(
         "Lower bound must be lower than the upper bound.");
   }
-  srand(seed);
+  std::random_device rd;
```
Mostah: If we use time(0) as default value in the random generator, we won't need this random device anymore !

test/learning/TestInitializeProfile.cpp
```diff
@@ -0,0 +1,180 @@
...
+                        given_perf));
+    }
+    perf_vect.push_back(vp);
+  }
+  PerformanceTable perf_tab = PerformanceTable(perf_vect);
+  AlternativesPerformance alt_perf = AlternativesPerformance(perf_tab, map);
+  ProfileInitializer profInit = ProfileInitializer(conf, alt_perf);
+  MRSortModel model = MRSortModel(3, 4);
+  profInit.initializeProfiles(model);
+  // EXPECT_TRUE(model.profiles.isProfileOrdered());
```
Mostah, line 182: You need to fix this test to make it work ! Find a way to test something :)

Mostah (approved):
We are almost there! A few nit and it should be good to go 👍 

Let's keep in mind that we need a bigger test (from input to output) for this algorithm but you can add it later if you prefer.

Mostah:
You need to remove the commented tests in TestMain before merging it

src/types/Profiles.cpp
thibmonsel: lets do logging after :)

include/utils.h
thibmonsel: This is still the same conceptually, std::mt19937 gen(seed); take as input a number and gives an integer.

Mostah: I know this is from a previous PR, but we should not set the seed to a default - constant - value, this has to be changed if we want to a random behaviour "by default"

src/learning/ProfileInitializer.cpp
```diff
@@ -199,16 +201,25 @@ std::vector<Perf> ProfileInitializer::initializeProfilePerformance(
...
-    // OPTIM : POSSIBILITY parallelization asynchrone
-    std::vector<Perf> p = ProfileInitializer::initializeProfilePerformance(
-        criterion, model.categories, catFreq);
-    perf_vec.push_back(Performance(p));
+  bool ordered = 0;
+  while (!ordered) {
+    std::vector<Performance> perf_vec;
+    Performance firstAltPerf = altPerformance_.getPerformanceTable()[0];
+    std::vector<std::string> criteriaIds = firstAltPerf.getCriterionIds();
+    std::vector<float> catFreq = ProfileInitializer::categoryFrequency();
```
Mostah: I think we should use `this->` here instead of `ProfileInitializer::`

src/learning/ProfileInitializer.cpp
```diff
@@ -199,16 +201,25 @@ std::vector<Perf> ProfileInitializer::initializeProfilePerformance(
...
+      std::vector<Perf> p = this->initializeProfilePerformance(
+          criterion, model.categories, catFreq);
+      perf_vec.push_back(Performance(p));
+    }
+    try {
+      Profiles p = Profiles(perf_vec, "crit");
+      model.profiles = p;
+      ordered = 1;
+    } catch (std::invalid_argument const &err) {
+      ordered = 0;
```
Mostah: nit: you do not need any instruction in the catch statement, as ordered is already at false by default

src/learning/ProfileInitializer.cpp
```diff
@@ -199,16 +201,25 @@ std::vector<Perf> ProfileInitializer::initializeProfilePerformance(
...
-  Performance firstAltPerf = altPerformance_.getPerformanceTable()[0];
-  std::vector<std::string> criteriaIds = firstAltPerf.getCriterionIds();
-  std::vector<float> catFreq = ProfileInitializer::categoryFrequency();
-  for (std::string criterion : criteriaIds) {
-    // OPTIM : POSSIBILITY parallelization asynchrone
-    std::vector<Perf> p = ProfileInitializer::initializeProfilePerformance(
-        criterion, model.categories, catFreq);
-    perf_vec.push_back(Performance(p));
+  bool ordered = 0;
+  while (!ordered) {
```
Mostah, line 206: Maybe, should we add a maximum iteration in the loop? Just to make sure we crash and return an error telling what's going on instead of just getting stuck in an infinite loop (if it ever happens)

src/types/PerformanceTable.cpp
```diff
@@ -164,36 +164,19 @@ void PerformanceTable::changeMode(std::string mode) {
...
-        if (index.count(p.getName()) > 0) {
-          // if alternative (or profile) already seen, add Perf to the right row
-          new_pt[index[p.getName()]].push_back(p);
-        } else {
-          // if alternative (or profile) already seen, add Perf to the right row
-          std::vector<Perf> v = {p};
-          new_pt.push_back(v);
-          index[p.getName()] = new_pt.size() - 1; // index of crit in the new_pt
-        }
+  for (std::vector<Perf> pv : pt_) {
```
Mostah: With this code you can only change mode from `alt` to `crit` no? We probably still need the otherway around still

test/types/TestDataGenerator.cpp
```diff
@@ -8,11 +8,11 @@
 #include <tuple>
 #include <utility>
 
-Config getTestConf() {
-  Config conf;
-  conf.data_dir = "../data/tests/";
-  return conf;
-}
+// Config getTestConf() {
```
Mostah: let's just get rid of it, it shouldn't be an issue

Mostah (commented):
Except my few comments (especially the change mode), it looks good to me!

src/learning/ProfileInitializer.cpp
thibmonsel: I added an if statement

include/types/Profiles.h
```diff
@@ -13,12 +13,16 @@
 class Profiles : public PerformanceTable {
 public:
   /**
-   * Profiles standard constructor (PerformanceTable surcharged)
+   * Profiles standard constructor (PerformanceTable surcharged). A Profile type
+   * is always represented in the mode_ = "crit" of a PerformanceTable. There is
+   * no such thing as a Profile in "alt" mode.
```
edix9, line 18: Is this still true (only one mode for profile)? 
If so, why do we have the choice of the mode for the profile?

include/utils.h
```diff
@@ -112,5 +118,20 @@ randomCategoriesLimits(int nbCategories, unsigned long int seed = time(NULL)) {
...
+ * Overloading << operator for std::unordered_map
+ *
+ * @param out ostream
+ * @param std::unordered_map  object
+ *
+ */
+template <typename K, typename V>
+std::ostream &operator<<(std::ostream &out, std::unordered_map<K, V> const &m) {
+  for (auto const &pair : m) {
+    out << "{, (" << pair.first << ": " << pair.second << "),";
```
edix9, line 131: Isn't the opening { displayed as many times as the for loop is running whereas the closing } is displayed only once?

src/types/Profiles.cpp
```diff
@@ -1,13 +1,18 @@
...
 #include <iostream>
+#include <random>
+#include <sstream>
 #include <string>
 #include <vector>
 
-Profiles::Profiles(std::vector<Performance> &perf_vect)
+Profiles::Profiles(std::vector<Performance> &perf_vect, std::string mode)
     : PerformanceTable(perf_vect) {
+  mode_ = mode;
```
edix9, line 15: Maybe we should have a check for the mode here (for example if the given mode is not a valid argument)

src/types/Profiles.cpp
```diff
@@ -16,23 +21,27 @@ Profiles::Profiles(std::vector<Performance> &perf_vect)
   }
 }
 
-Profiles::Profiles(int nb_of_prof, Criteria &crits, std::string prefix)
+Profiles::Profiles(int nb_of_prof, Criteria &crits, std::string mode,
+                   std::string prefix)
     : PerformanceTable(nb_of_prof, crits, prefix) {
   this->generateRandomPerfValues();
+  mode_ = mode;
```
edix9, line 28: same here (check on the mode?)

test/types/TestAlternativesPerformance.cpp
```diff
@@ -17,7 +17,7 @@ TEST(TestAlternativesPerformance, TestConstructorBaseConstructor) {
       os.str(),
       "AlternativesPerformance( PerformanceTable[ Performance: Perf( name : "
       "a0, crit : crit0, value : 0 ) Perf( name : a0, crit : crit1, value : 0 "
-      ") | ], AlternativesAssignment{ a0->Category(id : , rank : -1) }");
+      ") | ], AlternativesAssignment{ {, (a0: Category(id : , rank : -1)),}");
```
edix9, line 20: I think this is related to the << operator in utils, but it is not easy to read here, and why is there random **,** at the begining and the end of the AA? 
Again, I think there are too many { for not enough } here

edix9 (commented):
Small comments but this looks good!!

PR #40, opened 2021-01-31
Thomas/linear solver and weight updater

Mostah:
WeightUpdater architecture

I - Context

The weight updater algorithm is responsible for updating the weight associated to each Criterion and the lambda (majority threshold) of the problem.
The weight updater algorithm will be divided in two main components (C++ classes): the LinearSolver and the WeightUpdater.

II - WeightUpdater

Attributes in args:
&Config - Global config of the application
LinearSolver - abstraction between the algorithm and the way we solve the linear problem
&AlternativesPerformance - input of the global problem

The weightUpdater will be responsible of computing the constraint matrix and pass it to the linear solver before updating the MRSortModel with the new weights and lambda.

III - LinearSolver

Attributes in args:
Delta: default 0.0001
&AlternativesPerformance - input of the global problem
&Config - Global config of the application

The linear solver make an interface between our algorithm and the way we solve the linear problem. Therefore, if at some point the way of solving the linear problem is changed (changing the library for example), no modification in the WeightUpdater class will be needed. 

The linear solver will take as input the constraint matrix and the variables of the problem to solve, will instantiate the solver model and compute the new weight and lambda.

Linear Problem:

Variables : 

w_j in [0,1] for j in {1, …, n_crit}
lambda in [0.5,1]
x_a in [0,1] for a in {1, …, n_alt}
y_a in [0,1] for a in {1, …, n_alt}
xp_a in [0,1] for a in {1, …, n_alt}
yp_a in [0,1] for a in {1, …, n_alt}

Constraints :

sum(w_j for j in {1, …, n_crit}) = 1
sum(w_j(a_i, b_h-1) if a_i>=bi_h-1) - x_a + xp_a = lambda, for a assigned to C_h, for h in [1, n_prof]
sum(w_j(a_i, b_h) if a_i>=bi_h) - y_a + yp_a = lambda - delta, for a assigned to C_h, for h in [0, n_prof-1]

Objective function

min(sum(x_ap+y_ap, for a in {0,…,n_alt}))

Constraint matrixes

Implementation with dimension (n_prof, n_alt, n_crit): vect<vect<vect<bool>>>
First dimension: each profile
Second dimension, in profile h: each row correspond to an alternative
Thirs dimension, in profile h, alt a: each element is the presence or absence of the criterion in the constraint. w0 + w3 + w4 is [true, false, false, true, true]


include/learning/LinearSolver.h
```diff
@@ -0,0 +1,137 @@
...
+   * problem for the x variables
+   * @param y_matrix matrix recapitulating the constraints to add to the linear
+   * problem for the y variables
+   *
+   */
+  void updateConstraints(std::vector<std::vector<std::vector<bool>>> x_matrix,
+                         std::vector<std::vector<std::vector<bool>>> y_matrix);
+
+  /** solve Solve the linear problem given the constraint matrix.
+   *
```
thibmonsel, line 46: solve twice

src/learning/LinearSolver.cpp
```diff
@@ -0,0 +1,224 @@
...
+    weights.push_back(solver->MakeNumVar(0., 1.0, "w" + std::to_string(i)));
+  }
+
+  // x, x' (xp), y, y' (yp) variables
+  for (int i = 0; i < ap.getNumberAlt(); i++) {
+    x_a.push_back(solver->MakeNumVar(0., 1.0, "x" + std::to_string(i)));
+    x_ap.push_back(solver->MakeNumVar(0., 1.0, "xp" + std::to_string(i)));
+    y_a.push_back(solver->MakeNumVar(0., 1.0, "y" + std::to_string(i)));
+    y_ap.push_back(solver->MakeNumVar(0., 1.0, "yp" + std::to_string(i)));
+  }
```
thibmonsel: xa, ya, yap, xap aren't they suppose to be positive reals (thesis page 71) ?

src/learning/LinearSolver.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+  // lambda variable
+  lambda = solver->MakeNumVar(0.5, 1.0, "lambda");
+
+  const double infinity = solver->infinity();
+  // sum weight = 1 constraint: -sum <= -1 and sum <= 1
+  // -sum <= -1
```
thibmonsel: I suppose that you do -sum <= -1 because the >= operator does not exist ?

src/learning/LinearSolver.cpp
```diff
@@ -0,0 +1,224 @@
...
+  for (operations_research::MPVariable *w_i : weights) {
+    w_p->SetCoefficient(w_i, -1);
+  }
+  weights_constraint.push_back(w_p);
+  // sum <= 1
+  auto w_m = solver->MakeRowConstraint(-infinity, 1, "weight_constraint_m");
+  for (operations_research::MPVariable *w_i : weights) {
+    w_m->SetCoefficient(w_i, 1);
+  }
+  weights_constraint.push_back(w_m);
```
thibmonsel: maybe name the variable w_m, w_p that are more meaningful ? w_p maybe w_maj (for majorant) and w_min (for minorant) ?

src/learning/LinearSolver.cpp
```diff
@@ -0,0 +1,224 @@
...
+        // cst_x_b2_a6_+ : - cst_x_b2_a6 <= 0
+        operations_research::MPConstraint *cst_m = solver->MakeRowConstraint(
+            -infinity, 0,
+            "cst_x_b" + std::to_string(h) + "_a" + std::to_string(alt) + "_+");
+
+        // cst_x_b2_a6_- : cst_x_b2_a6 <= 0
+        operations_research::MPConstraint *cst_p = solver->MakeRowConstraint(
+            -infinity, 0,
+            "cst_x_b" + std::to_string(h) + "_a" + std::to_string(alt) + "_-");
+
```
thibmonsel: I like the names great for debugging !

src/learning/LinearSolver.cpp
```diff
@@ -0,0 +1,224 @@
...
+LinearSolver::solve(std::vector<std::vector<std::vector<bool>>> x_matrix,
+                    std::vector<std::vector<std::vector<bool>>> y_matrix) {
+  this->updateConstraints(x_matrix, y_matrix);
+  const operations_research::MPSolver::ResultStatus result_status =
+      solver->Solve();
+  if (result_status != operations_research::MPSolver::OPTIMAL) {
+    conf.logger->warn("Solver couldn't find an optimal solution.");
+    throw std::logic_error("Solver couldn't find an optimal solution");
+  }
+  std::ostringstream ss;
```
thibmonsel, line 214: should we do a timeout if the solver takes too long ?

src/learning/WeightUpdater.cpp
```diff
@@ -0,0 +1,112 @@
...
+#include <sstream>
+
+WeightUpdater::WeightUpdater(AlternativesPerformance &ap, Config &conf)
+    : ap(ap), conf(conf), solver(ap, conf) {}
+
+WeightUpdater::WeightUpdater(const WeightUpdater &wu)
+    : solver(wu.solver), ap(wu.ap), conf(wu.conf) {}
+
+WeightUpdater::~WeightUpdater() {}
+
```
thibmonsel, line 13: Will the linear solver delete itself here ?

src/learning/WeightUpdater.cpp
```diff
@@ -0,0 +1,112 @@
...
+  std::vector<std::vector<std::vector<bool>>> x_matrix;
+  auto profs_pt = model.profiles.getPerformanceTable();
+  auto alts_pt = ap.getPerformanceTable();
+  auto alts_assign = ap.getAlternativesAssignments();
+  // As the profiles are stored in descending order (h = 0 is the highest
+  // profile, which is not the case in the theorical formulation), going in
+  // reverse.
+  for (int h = profs_pt.size() - 2; h >= 0; h--) {
+    std::vector<std::vector<bool>> x_h;
```
thibmonsel: Isn't category 0 suppose to be the worst ? so h=0 is the profile delimiting cat0 and cat1. 
As of today profiles are set row by row with {b0,b1,b2,...}

src/learning/WeightUpdater.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+  std::vector<std::vector<std::vector<bool>>> y_matrix;
+  auto profs_pt = model.profiles.getPerformanceTable();
+  auto alts_pt = ap.getPerformanceTable();
+  auto alts_assign = ap.getAlternativesAssignments();
+
+  // As the profiles are stored in descending order (h = 0 is the highest
+  // profile, which is not the case in the theorical formulation), going in
+  // reverse.
+  for (int h = profs_pt.size() - 1; h > 0; h--) {
```
thibmonsel: same comment here

src/learning/WeightUpdater.cpp
```diff
@@ -0,0 +1,112 @@
...
+      std::vector<bool> y_h_alt;
+      // if alt is assigned to category h (otherwise, append empty vector)
+      if (ap.getAlternativeAssignment(alt[0].getName()).getCategoryId() ==
+          profs_pt[h][0].getName()) {
+        for (int j = 0; j < alt.size(); j++) {
+          // condition: aj >= bj_h
+          y_h_alt.push_back(alt[j].getValue() >= profs_pt[h][j].getValue());
+        }
+      }
+      y_h.push_back(y_h_alt);
```
thibmonsel, line 84: variable names are really explicit that good !

src/learning/WeightUpdater.cpp
```diff
@@ -0,0 +1,112 @@
...
+  }
+  return y_matrix;
+}
+
+bool WeightUpdater::modelCheck(MRSortModel &model) {
+  if (model.profiles.getNumberCrit() != ap.getNumberCrit()) {
+    return false;
+  }
+  // both are supposed to be in mode alt
+  auto profs = model.profiles.getPerformanceTable();
```
thibmonsel, line 103: maybe get a throw statement by checking the mode

test/learning/TestLinearSolver.cpp
```diff
@@ -0,0 +1,171 @@
...
+  Criteria crits = Criteria(2);
+  Config conf;
+  AlternativesPerformance ap = AlternativesPerformance(4, crits);
+  LinearSolver ls = LinearSolver(ap, conf);
+  ls.initializeSolver();
+
+  operations_research::MPSolver *solver = ls.getSolver();
+  EXPECT_EQ(solver->NumVariables(), 19);
+  EXPECT_EQ(solver->NumConstraints(), 2);
+}
```
thibmonsel: maybe explicitly give the addition that gives 19 variables

test/learning/TestWeightUpdater.cpp
```diff
@@ -0,0 +1,99 @@
...
+  std::vector<float> vec0 = {0.8, 0.6}; // good candidate
+  std::vector<float> vec1 = {1, 0.5};   // good candidate
+  std::vector<float> vec2 = {0.2, 0.4}; // bad candidate
+  perf_vect.push_back(Performance(crits, vec0, "a0"));
+  perf_vect.push_back(Performance(crits, vec1, "a1"));
+  perf_vect.push_back(Performance(crits, vec2, "a2"));
+  Category cat0 = Category("cat0", 0); // bad category
+  Category cat1 = Category("cat1", 1); // good category
+
+  std::unordered_map<std::string, Category> map =
```
thibmonsel: This will have to change will the disappearance of Performance.

> u will have to include utils in the test file
> change Performance with createVectorPerf("a0, cries, vec0)

test/learning/TestWeightUpdater.cpp
```diff
@@ -0,0 +1,99 @@
...
+  Criteria crits = Criteria(2, "crit");
+
+  std::vector<float> vec1 = {0.7, 0.5};
+  std::vector<float> vec0 = {0.3, 0.4};
+  std::vector<Performance> perf_vect;
+  perf_vect.push_back(Performance(crits, vec1, "cat1"));
+  perf_vect.push_back(Performance(crits, vec0, "cat0"));
+  Profiles profs = Profiles(perf_vect);
+
+  MRSortModel model = MRSortModel(crits, profs, cats, 0.6);
```
thibmonsel: same

thibmonsel (approved):

include/learning/LinearSolver.h
Mostah: Yeah, name of functon and then its description 🤣

src/learning/LinearSolver.cpp
Mostah: Indeed that's a huge mistake from a copy paste 👍

src/learning/LinearSolver.cpp
Mostah: Exactly, the solver can only take constraints with <= operator

src/learning/LinearSolver.cpp
Mostah: You are right, they are even confusing for me

src/learning/LinearSolver.cpp
Mostah: It is handled by the solver natively, that could be the cause of no optimal solution found !

src/learning/WeightUpdater.cpp
Mostah: Yes, the LinearSolver class is stored by value, thus it will be deleted when the WeightUpdater is deleted. The ortools linear solver being stored by pointer has to be deleted manually, that's why we have it in the LinearSolver destructor

src/learning/WeightUpdater.cpp
Mostah: No, for the folowing profile:
```
    cat0  cat1
b1   0.8   0.6
b0   0.4   0.3
```

in the way we defined the Profiles performance table it would be:
```
prof = [[0.8,0.6],[0.4,0.3]]
```
thus b0 is actually prof[1] and not prof[0].

That's why we are going in reverse if we compare with the thesis

test/learning/TestWeightUpdater.cpp
Mostah: Will do after yours is merged!

include/learning/LinearSolver.h
```diff
@@ -0,0 +1,137 @@
...
+  Config getConf() const;
+
+private:
+  AlternativesPerformance &ap;
+  operations_research::MPSolver *solver;
+  std::string solver_name;
+  Config &conf;
+
+  // value use to transform strict inequalities into non-strict ones
+  float delta;
```
edix9, line 121: We might align on what we call this. I called it epsilon on ProfileUpdater since it is more like a small parameter to ensure the order is maintained.

src/learning/LinearSolver.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+  // reset variables vectors
+  x_a.clear();
+  x_ap.clear();
+  y_a.clear();
+  y_ap.clear();
+  weights.clear();
+
+  // reset previous solver
+  solver->Clear();
```
edix9, line 65: why is this one capitalized? (C maj) 
I'm just curious

src/learning/LinearSolver.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+  const double infinity = solver->infinity();
```
edix9, line 67: what is the purpose of infinity here? I don't really understand, maybe you could just add a comment to explain?

Mostah: It's on the thesis equations, it's use to specify that the variables can be in the range until infinity

src/learning/LinearSolver.cpp
Mostah: This is the convention used by ortools, `solver->Clear` is not something we implemented ourself

include/learning/LinearSolver.h
Mostah: I will stick with delta here as it is the name of the variable use in the thesis :)

PR #41, opened 2021-02-03
Thibault/remove performance

thibmonsel:
It is branched off the ProfileInitializer branch.

Took out all the Performance type and added 2 functions in utils to create vector of Perf.
All tests run fine but I wouldn't be surprised that we would need to go back to change some small things later. 

Enjoy


Mostah (approved):
Nice work ! It looks good to me :)

PR #44, opened 2021-02-25
install matplotlib cpp with an example

Mostah:


thibmonsel (approved):

thibmonsel:
I found a better library for plotting since Matplotlib was a very small version of its python equivalent.
The package is : https://github.com/alandefreitas/matplotplusplus
It is supported with the Gnuplot package and all other installation are in the thibault/importingData2 branch.

PR #46, opened 2021-03-10
fixed changeMode issue with Profiles

thibmonsel:
This function add mandatory attribute mode for PT.
Fixed changeMode bug with Profiles object.


src/types/PerformanceTable.cpp
```diff
@@ -395,4 +427,23 @@ void PerformanceTable::display() {
...
     }
   }
-}
\ No newline at end of file
+}
+
+bool PerformanceTable::operator==(const PerformanceTable &pt) const {
+
+  if (this->pt_.size() != pt.getPerformanceTable().size() &&
+      this->pt_[0].size() != pt.getPerformanceTable().size()) {
```
Mostah: shouldn't the check here be `pt_.size != pt.size || pt_[0].size != pt[0].size` ?

test/learning/TestInitializeProfile.cpp
```diff
@@ -175,8 +175,16 @@ TEST(TestProfileInitializer, TestInitializeProfiles) {
   ProfileInitializer profInit = ProfileInitializer(conf, alt_perf);
   MRSortModel model = MRSortModel(3, 4);
   profInit.initializeProfiles(model);
+  std::cout << model.profiles << std::endl;
```
Mostah: You probably forgot to remove the `std::cout` from your test files

test/types/TestChangeMode.cpp
```diff
@@ -0,0 +1,79 @@
...
+#include "../../include/types/PerformanceTable.h"
+#include "../../include/utils.h"
+
+#include "gtest/gtest.h"
+#include <sstream>
+#include <utility>
+
+// Checking PerformanceTable changeModes
+
+TEST(TestPerformanceTable, TestChangeMode1) {
```
Mostah: If you don't execute any test in here, best remove it :)

src/types/PerformanceTable.cpp
thibmonsel: you have to call getPerformanceTable() for pt.

test/types/TestChangeMode.cpp
thibmonsel: Yup didn't see I added it.

src/types/PerformanceTable.cpp
Mostah: The comment was about the logic (&& -> ||) and I believe you forgot a [0] in the second part

thibmonsel: yup thanks for that I got tired probably ahah.

Mostah (approved):

PR #47, opened 2021-03-12
Emma/profile updater

edix9:
[Profile Updater.docx](https://github.com/Mostah/fastPL/files/6128845/Profile.Updater.docx)
**Profile Updater:**
Profile Updater algorithm is responsible for updating the profiles given the fixed weight and lambda. It will evaluate each move of each profile on each criterion to see what moves have the best impact on the alternatives assignments.

**_Compute above desirability:_**
Checking all the possible moves of a profile b between b and b_above on one criterion, b is moving up.
b represents the border between cat (below b) and cat_above (above b).
For each alternative alt between b and b_above on criterion crit:
- Compute the impact if b moves right above alt (and before the following alt). There are 5 eventualities for this move (Q, R, W, V, T, see thesis Chap 3.3.4 Variant 2).
- Wright the desirability in a dictionary of the possible moves (desirability_above). key = new performance of the profile, value = desirability ( P(b_j^h+δ) in thesis). δ in thesis corresponds to the difference between b current perf and its new perf.
epsilon ensures that the profile moves right above the alternative and not at the same place, so the classifications (alternatives are either above or below a profile but not equal).
returns the dictionary of desirabilities.

**_Compute below desirability:_** 
Checking all the possible moves of a profile b between b and b_below on one criterion, b is moving down.
b represents the border between cat (below b) and cat_above (above b).
epsilon ensures that the profile moves right below the alternative and not at the same place, so the classifications (alternatives are either above or below a profile but not equal).
returns the dictionary of desirabilities.

**_Chose Max Desirability:_**
In the map of all moves for b between b_below and b_above, select the move that has the best impact for the alternative’s classification. 

**_Update Tables:_**
Updates tables and algorithm’s data with the new profile value:
- concordance table
- good assignment count
- alternative assignments
- profiles

**_Optimize Profile:_**
Optimizes one profile using the profileUpdater methods:
	Computes the desirability of each move of the profile on each criterion
	If the desirability of a move for one criteria is above a random float r[0,1], then we move the profile
	Update the tables accordingly

**_Optimize:_**
For each profile, run the optimize profile method. Returns the ratio of (good assignments)/(total assignments)


src/types/MRSortModel.cpp
```diff
@@ -37,52 +37,85 @@ MRSortModel::MRSortModel(int n_cat, int n_crit, std::string id)
...
 }
 
-AlternativesPerformance MRSortModel::categoryAssignment(PerformanceTable &pt) {
+Category
+MRSortModel::categoryAssignment(std::vector<Perf> &alt,
+                                std::vector<std::vector<Perf>> &profiles_pt) {
+  bool assigned = false;
+  Category cat_assignment;
+  // For all alt, looping over all profiles in descending order
+  for (int h = 0; h < profiles_pt.size(); h++) {
```
Mostah: Let's simplify the loop here: instead of using `prof_index`, just loop in descending order from `profiles_pt.size() - 1` to 0

src/types/PerformanceTable.cpp
```diff
@@ -220,36 +220,45 @@ void PerformanceTable::sort(std::string mode) {
...
-  if (lower_b != pv.end() && upper_b != pv.end()) {
-    return subVector(pv, std::distance(pv.begin(), lower_b),
-                     std::distance(pv.begin(), upper_b) - 1);
-  } else if (upper_b == pv.end()) {
-    return subVector(pv, std::distance(pv.begin(), lower_b), pv.size() - 1);
-  } else {
-    // return empty perf vect
-    std::vector<Perf> v;
-    return v;
+  std::vector<Perf> v;
```
Mostah, line 261: This function is a bit confusing now as in the `crit` mode we assume the table to be sorted to perform a slice and in the `alt` mode we don't assume anything and just look for the matches by going over all possibilities. Could we rewrite this to: either keep the two modes with the same implementation and assumptions, or remove the unused mode ?

src/types/Profiles.cpp
```diff
@@ -90,4 +90,101 @@ bool Profiles::isProfileOrdered() {
...
+        }
+      }
+    }
+    throw std::invalid_argument("Profile not found.");
+  } else {
+    throw std::domain_error("Profiles perftable mode corrupted.");
+  }
+}
+
+void Profiles::setPerf(std::string name, std::string crit, float value) {
```
Mostah, line 134: I think your implementation here could be simplified, take a look at getPerf: https://github.com/Mostah/fastPL/blob/master/src/types/PerformanceTable.cpp#L119
As you need to write something in the table, you'll have to use `&` :

```
for (std::vector<Perf> &p : pt_) {
        for (Perf &perf : p) {
            ...
          }
        }
```

test/types/TestMRSortModel.cpp
```diff
@@ -52,29 +52,82 @@ TEST(TestMRSortModel, TestBaseConstructorError) {
...
+  EXPECT_FLOAT_EQ(c_alt1_b2, 0.3);
+
+  // TEST COMPUTE CONCORDANCE TABLE
+  std::unordered_map<std::string, std::unordered_map<std::string, float>> ct =
+      mrsort.computeConcordanceTable(pt_);
+
+  EXPECT_FLOAT_EQ(ct["b0"]["alt0"], 0.3);
+  EXPECT_FLOAT_EQ(ct["b2"]["alt1"], 0.3);
+
+  // Display ct
```
Mostah: Don't forget to remove this before merging the branch :)

test/types/TestPerformanceTable.cpp
```diff
@@ -254,15 +254,16 @@ TEST(TestPerformanceTable, TestGetAltBetween) {
...
-  try {
-    perf_table.getAltBetween("crit0", 0, 1);
-    FAIL() << "should have throw domain error.";
-  } catch (std::domain_error const &err) {
-    EXPECT_EQ(err.what(),
-              std::string("Mode must be crit but is currently set to alt."));
-  } catch (...) {
-    FAIL() << "should have throw domain error.";
-  }
+  // try {
```
Mostah: same here, if this is of no use anymore !

include/types/ProfileUpdater.h
```diff
@@ -0,0 +1,143 @@
...
+#include "MRSortModel.h"
+
+class ProfileUpdater {
+public:
+  /**
+   * ProfileUpdater standard constructor
+   *
+   * @param epsilon
+   */
+  ProfileUpdater(AlternativesPerformance &altPerf_data,
```
Mostah: I see you are not really using it, but can we add the config as well here? It'll make things more consistent with the other learning classes and if we need it for later it will already be there !

src/types/ProfileUpdater.cpp
```diff
@@ -0,0 +1,352 @@
...
+#include <iostream>
+#include <map>
+#include <string>
+#include <typeinfo>
+
+ProfileUpdater::ProfileUpdater(AlternativesPerformance &altPerf_data,
+                               float epsilon)
+    : altPerf_data(altPerf_data), epsilon_(epsilon) {}
+
+ProfileUpdater::ProfileUpdater(const ProfileUpdater &profUp)
```
Mostah: To entirely copy the state of the ProfileUpdater, you should also set the new value of `good_` in here!

src/types/ProfileUpdater.cpp
```diff
@@ -0,0 +1,352 @@
...
+
+  // Initializing the map of desirability indexes
+  std::unordered_map<float, float> desirability_above;
+  float numerator = 0;
+  float denominator = 0;
+
+  for (Perf alt : alt_between) {
+    // Checking if the move will not go above b_above
+    if (alt.getValue() + epsilon < b_above.getValue()) {
+      std::cout << "IN THE LOOP FOR " << alt.getName() << std::endl;
```
Mostah: Before merging the branch, don't forget to remove all you debugging `std::cout` ! If you want to keep them, consider using logs in `debug` level (they will only be run while developping)

src/types/ProfileUpdater.cpp
```diff
@@ -0,0 +1,352 @@
...
+
+    // Update concordance table
+    float c = ct[b_old.getName()][alt.getName()] + w;
+    ct[b_old.getName()][alt.getName()] = c;
+    // Update profile
+    model.profiles.setPerf(b_new.getName(), b_new.getCrit(), b_new.getValue());
+
+    // New assignment
+    altPerf_model.changeMode("alt");
+    auto alternative = altPerf_model.operator[](alt.getName());
```
Mostah: isn't `alternative` the same as your `alt` variable from the loop here? Also, keep in mind that the `[]` operator for performance tables is inefficient (O(size of the table))

src/types/ProfileUpdater.cpp
```diff
@@ -0,0 +1,352 @@
...
+float ProfileUpdater::optimize(
+    MRSortModel &model,
+    std::unordered_map<std::string, std::unordered_map<std::string, float>> &ct,
+    AlternativesPerformance &altPerf_model) {
+  int i = 0;
+  for (std::vector<Perf> profile : model.profiles.getPerformanceTable()) {
+    if (model.profiles.getMode() != "alt") {
+      model.profiles.changeMode("alt");
+      if (!model.profiles.isProfileOrdered()) {
+        model.profiles.sort("alt");
```
Mostah: You might have to double check here, I don't believe the sort method order in a profile order

include/types/ProfileUpdater.h
```diff
@@ -0,0 +1,143 @@
...
+   */
+  float optimize(MRSortModel &model,
+                 std::unordered_map<std::string,
+                                    std::unordered_map<std::string, float>> &ct,
+                 AlternativesPerformance &altPerf_model);
+
+private:
+  float epsilon_;
+  AlternativesPerformance &altPerf_data;
+  int good_;
```
Mostah: After reading your code, I still don't understand the purpose of this `good_` variable ! It looks like this doesn't have to be part of the state of this object and could be defined and store in your functions instead of here, but correct me if I'm wrong !

src/types/ProfileUpdater.cpp
```diff
@@ -0,0 +1,352 @@
...
+  }
+}
+
+float ProfileUpdater::optimize(
+    MRSortModel &model,
+    std::unordered_map<std::string, std::unordered_map<std::string, float>> &ct,
+    AlternativesPerformance &altPerf_model) {
+  int i = 0;
+  for (std::vector<Perf> profile : model.profiles.getPerformanceTable()) {
+    if (model.profiles.getMode() != "alt") {
```
Mostah: Shouldn't these checks be before the for loop ? Or do you expect the profiles to change mode in between loop?

src/types/ProfileUpdater.cpp
```diff
@@ -0,0 +1,352 @@
...
+        model.profiles.sort("alt");
+      }
+    }
+    Category cat_below = model.categories.getCategoryOfRank(i);
+    Category cat_above = model.categories.getCategoryOfRank(i + 1);
+    this->optimizeProfile(profile, cat_below, cat_above, model, ct,
+                          altPerf_model);
+  }
+  int n_alt = altPerf_data.getNumberAlt();
+  return (good_ / n_alt);
```
Mostah: The accuracy will be computed a bit further in the algorithm, you don't have to perform this here!

src/types/ProfileUpdater.cpp
```diff
@@ -0,0 +1,352 @@
...
+}
+
```
Mostah: You probably haven't finish yet but just a reminder: you must have a method that only take as an arguments a `MRSortModel &model` model

test/types/TestProfileUpdater.cpp
```diff
@@ -0,0 +1,221 @@
...
+
+  std::unordered_map<std::string, Category> assignment;
+  assignment["alt0"] = categories.getCategoryOfRank(0);
+  assignment["alt1"] = categories.getCategoryOfRank(0);
+  assignment["alt2"] = categories.getCategoryOfRank(1);
+  assignment["alt3"] = categories.getCategoryOfRank(0);
+  return AlternativesPerformance(perf_vect, assignment);
+}
+
+TEST(TestProfileUpdater, TestBaseConstructor) {
```
Mostah: You don't have to add a test for the constructor if there is nothing to test about it !

Mostah (commented):
There is a few things, but overall this is high quality code ! Congrats on your work, keep it up we are almost there 😄

include/types/Profiles.h
```diff
@@ -74,6 +74,26 @@ class Profiles : public PerformanceTable {
...
+  getBelowAndAboveProfile(std::string profName);
+
+  /**
+   * setPerf set a Perf given the alt name, crit and value
+   *
+   * @param name name of the alt or profile we want to update
+   * @param crit name of the crit we want to update
+   * @param value new perf value
+   */
+  void setPerf(std::string name, std::string crit, float value);
```
thibmonsel: Should we better see those types of functions be put in PerformanceTable than Profiles ?

src/types/ProfileUpdater.cpp
```diff
@@ -0,0 +1,352 @@
...
+    std::cout << key_max << std::endl;
+  }
+}
+
+float ProfileUpdater::optimize(
+    MRSortModel &model,
+    std::unordered_map<std::string, std::unordered_map<std::string, float>> &ct,
+    AlternativesPerformance &altPerf_model) {
+  int i = 0;
+  for (std::vector<Perf> profile : model.profiles.getPerformanceTable()) {
```
thibmonsel: In this function i is always set to 0 ?

thibmonsel: Maybe you made test for only 2 categorie, idk ?

src/types/Profiles.cpp
```diff
@@ -90,4 +90,101 @@ bool Profiles::isProfileOrdered() {
...
+          below = pt_[h - 1];
+          above = pt_[h + 1];
+          return std::make_pair(below, above);
+        }
+      }
+    }
+    throw std::invalid_argument("Profile not found.");
+  } else {
+    throw std::domain_error("Profiles perftable mode corrupted.");
+  }
```
thibmonsel: made the error message here is miss leading, probably refer to the error as a "crit" mode ?

src/types/Profiles.cpp
thibmonsel: This will boost code readability and memory efficiency with &

thibmonsel (approved):
Code looks great :), open to maybe moving the Profiles method to PT if that makes sense to everybody in order to keep separate conceptual concepts that make up PT and Profiles objects.

include/types/Profiles.h
Mostah: I actually prefer it in here instead of in PerformanceTable, but I'm open to discussion! The reason is, setPerf is highly uneficient and we should avoir using it. However - correct me if I'm wrong - Profiles object should remain quite small in size, in comparison to datasets. Therefore, only defining it for Profiles prevent its use for datasets. 

Anyway, it could be great to add a comment on the doc that this is eneficient and shouldn't be use often

src/types/PerformanceTable.cpp
edix9: I agree, not sure what should be best though: 
- in ProfileUpdater the function is used in `crit` mode and probably in `alt` mode too (the last one depends on the altPerf table given by the model). So not sure if we can remove an unused mode.
- the way of performing the search might by faster on a single core with the check if the table is sorted, however if parallelized it might just be easier without the check for sorted. What do you think?

Mostah: If you think both modes are usefull, I'd rather implement two different functions than having two different behaviour given the mode: one with the assumption of a sorted pt `getAltBetweenSorted` for ex and one without `getAltBetween`. This would make things more readable !

I agree that with parallelization the sorted method might be obsolete, but for the moment I think we should focus on having satisfying performances on a single core! However the idea of sorting the pt before executing altBetween might not be really optimized (I haven't think that through), so it's up to you to do whatever seems best in this situation :)

include/types/ProfileUpdater.h
edix9: Actually the good variable shouldn't be in profile updater as the goal if this is to measure the right assigments and therefore select the best models. It might be useful to have is declared in the pipeline? However we need to thing of how we are going to count the right assignments at the end of a model improvement.

src/types/ProfileUpdater.cpp
edix9: You're right, my bad!

src/types/PerformanceTable.cpp
edix9: I can make two functions however I'm not sure how to implement the "alt" part for the sorted method in mirror of the "crit" implementation here...

include/types/Profiles.h
thibmonsel: sounds good

include/types/ProfileUpdater.h
thibmonsel: we can also have a utils function that computes the CA like in PYMCDA

src/types/PerformanceTable.cpp
Mostah: We do not have to implement it if we don't intend to use it, just trow an error if the table is not in the mode the function has been implemented for

include/types/ProfileUpdater.h
Mostah: Yes, that'll be implemented in the HeuristicPipeline

edix9: I'll remove the good attribute from the class then, and we'll have the computation of the accuracy elsewhere. 
We should keep in mind that it might be useful to have the accuracy as an attribute from the model, so we don't need to recalculate the good/bad assignments at the end but we can update it directly in the ProfileUpdater.
The computation of good and bad assignment is already done in the ProfileUpdater it might loose efficiency to compute it again instead of just updating the value directly.

Mostah: Okay, let's add the equivalent of a score attribute in the model class and update its value in your algorithm then !

edix9: Done, but we will need to initialize the value in the pipeline before going through the profile updater, alright?

Mostah: What do you mean? Can't the value be initialize at zero at the creation of the model?

edix9: yes but at before going through the profile updater the score will not be 0 (unless it's a very bad model)

include/learning/ProfileUpdater.h
```diff
@@ -6,7 +6,18 @@
 #include <vector>
```
Mostah, line 13: Can you add some doc at the file level? Something like 
```
/**
 * @file ProfileUpdater.h
 * @brief Profile updater algorithm.
 *
 */
```

Mostah (approved):
Things look good to me as is ! We could merge this and see what happens on the heuristic pipeline :)

PR #52, opened 2021-03-31
Emma/debug prof updater

edix9:
- corrected function choose_max_des that returned only one vaue instead of the two necessary
- corrected random float generator in ProfUpdater
- corrected comparison of rank instead of id for categories which results in ProfUpdater working well!
- some small updated on README


LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2021 Thomas Vindard
```
thibmonsel: Don't know how Licenses work but need all Copyright collaborator names ? idk

thibmonsel (approved):

LICENSE
Mostah: Don't know why this is here as it is already in the repo! According to the MIT Licence, the name should be the person or the organisation hosting the repository so as it is my account I only put my name... I have no right on the code with this licence don't worry ;)

Will check with Laurent

Mostah (approved):

PR #54, opened 2021-04-01
move Perf args to public and remove getter / setters

Mostah:
Getter and setters seemed to take quite a while in the profiling for Perf and Category, therefore I moved all args for these classes to public and removed the getter / setters of these methods.

Also removed unecessary copy of Perf in the ProfileInitialization algorithm


thibmonsel (approved):
all heroes don't wear capes 👯

PR #55, opened 2021-04-07
Thibault/init optim

thibmonsel:
change some small code and added comments for optimisation


src/learning/ProfileInitializer.cpp
```diff
@@ -95,10 +95,11 @@ std::vector<Perf> ProfileInitializer::getProfilePerformanceCandidates(
     catAbove = catBelow + 1;
   }
 
+  std::unordered_map<std::string, Category> assig = altPerformance_.getAlternativesAssignments();
```
Mostah, line 98: If I remember correctly, the issue comes from WeightedProbability, line 142 & 149

